PR #29, opened 2015-06-16
invalid AS or prefix length

smlng:
This implements more specific validation result to distinguish between invalid AS or invalid prefix length, if any ROA for the IP prefix exists in the RPKI database at all.


smlng:
Use reason in client tools to distinguish invalid AS and prefix length, no modification of rtrlib required!

PR #30, opened 2015-06-16
enhanced cli-validator to distinguish invalid AS and invalid prefix length

smlng:


smlng:
Not needed, evaluate (optional) reason to distinguish between invalid AS and invalid length. This should be done by a third-party tools using cli-validator.

PR #32, opened 2015-10-21
Fix freeing memory

shaddeus:


fho:
Thanks a lot @shaddeus, both commits were applied on master.

PR #34, opened 2015-10-21
SSH Transport: use new LibSSH API

shaddeus:
The old version did not catch an interrupt signal in the rtr_fsm thread for closing ssh socket.

### DIFF
--- a/README
+++ b/README
@@ -18,7 +18,7 @@ Requirements
 To build the RTRlib, the CMake build system must be installed.
 
 To establish an SSH connection between RTR-Client and RTR-Server, the
-libssh 0.6.x library must also be installed.
+libssh 0.6.x or higher library must also be installed.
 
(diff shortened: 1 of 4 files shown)


shaddeus:
I split commit into smaller commits.

libssh 0.6.0 (https://www.libssh.org/2014/01/08/libssh-0-6-0/) fix timeout handling in function ssh_channel_select(). The function ssh_channel_select() unlike ssh_channel_poll() can catch signals, so rtrlib can stop ssh transport using signal SIGUSR1.

Is it OK?

shaddeus:
The API prefix ```ssh_``` was added in version 0.5.0 already:
https://www.libssh.org/2011/06/01/libssh-0-5-0/

shaddeus:
Great, thanks!

PR #35, opened 2015-10-21
RTRLib Manager: Reverse finding next group by preference while RTR_ERROR_

shaddeus:
At first should try the low (more) preferenced caches and then the high (less).


fho:
thanks! applied commit on master branch

PR #36, opened 2015-10-21
Remove collisional namespace

shaddeus:
Probably should be beneficial for integrators of RTRLib.

Renaming:
```
struct ip_addr  -> struct rtr_ip_addr
enum ip_version -> rtr_ip_version
IPV6            -> RTRLIB_IPV6
IPV4            -> RTRLIB_IPV4
```

A few symbols were collisional with BIRD namespace, however BIRD will link RTRLib dynamically using dlopen().

### DIFF
--- a/doxygen/examples/rtr_mgr.c
+++ b/doxygen/examples/rtr_mgr.c
@@ -76,8 +76,8 @@ int main(){
         sleep(1);
 
     //validate the BGP-Route 10.10.0.0/24, origin ASN: 12345
-    struct rtr_ip_addr pref;
-    ip_str_to_addr("10.10.0.0", &pref);
+    struct lrtr_ip_addr pref;
+    lrtr_ip_str_to_addr("10.10.0.0", &pref);
     enum pfxv_state result;
     rtr_mgr_validate(conf, 12345, &pref, 24, &result);
 
--- a/doxygen/examples/rtr_mgr.c
+++ b/doxygen/examples/rtr_mgr.c
@@ -76,7 +76,7 @@ int main(){
         sleep(1);
 
     //validate the BGP-Route 10.10.0.0/24, origin ASN: 12345
-    struct ip_addr pref;
+    struct rtr_ip_addr pref;
     ip_str_to_addr("10.10.0.0", &pref);
     enum pfxv_state result;
     rtr_mgr_validate(conf, 12345, &pref, 24, &result);
--- a/rtrlib/lib/ip.c
+++ b/rtrlib/lib/ip.c
@@ -11,9 +11,9 @@
 #include <string.h>
 #include "rtrlib/lib/ip.h"
 
-bool ip_addr_is_zero(const struct ip_addr prefix)
+bool ip_addr_is_zero(const struct rtr_ip_addr prefix)
 {
-    if(prefix.ver == IPV6) {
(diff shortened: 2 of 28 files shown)


fho:
Thanks, I agree the lib/ functions should be prefixed to avoid nameclashes.

I see the following problems with your patch:
- It makes the naming more inconsistent. To keep the naming consistent, all functions should be prefixed with the same String in a header file. Could you also add prefixed to the other definitions in lib?
- The prefix "rtr_" is already used for the rtrlib/rtr component. 
The most functions are related to the transport. "tr_" seems to fit better. The utility functions related ot the transport could also be moved to rtrlib/transport

shaddeus:
Would it be alright if I change in the last commit this
```
ip_addr    -> rtrlib_ip_addr
ip_version -> rtrlib_ip_version
```
and add this?
```
ipv4_addr -> rtrlib_ipv4_addr
ipv6_addr -> rtrlib_ipv6_addr
```

waehlisch:
There is heavy push to merge the Quagga extension into the Quagga main branch. I think it would be better fo finish this API discussion before the merge. @fho comments?

reuteran:
I think adding the prefix 'rtr_' for all ip*, ipv4*, and ipv6* functions and structs would be OK, since 'rtr_ip' is still a distinct prefix from just 'rtr_'. Using tr_ could be confusing since its rtrlib specific jargon, but the rtr_ip prefix makes it clear that this belongs to rtrlib (or at least something that implements the rtr protocol). To avoid confusion we could also use 'lrtr_ip' (for librtr), but Im not sure thats much better. @waehlisch @fho thoughts?

waehlisch:
I agree that the prefix 'tr_' is confusing. I tend to vote either for 'rtr_' or 'lrtr_'.

waehlisch:
@shaddeus, we discussed this again and actually prefer the prefix 'lrtr_' as this gives a bit more context wrt the directory without making it too complex. Can you change your PR? Sorry for the additional load.

shaddeus:
Well done. Do you want something like this?
```
struct ip_addr     ->  struct lrtr_ip_addr
  enum ip_version  ->    enum lrtr_ip_version
       ipv4_addr   ->         lrtr_ipv4_addr
       ipv6_addr   ->         lrtr_ipv6_addr
       IPV6        ->         RTRLIB_IPV6
       IPV4        ->         RTRLIB_IPV4
```
or will you be satisfied only with?
```
struct ip_addr     ->  struct lrtr_ip_addr
  enum ip_version  ->    enum lrtr_ip_version
       IPV6        ->         RTRLIB_IPV6
       IPV4        ->         RTRLIB_IPV4
```

waehlisch:
First option (see @fho comment "To keep the naming consistent, all functions should be prefixed with the same String in a header file. Could you also add prefixed to the other definitions in lib?").

shaddeus:
I added prefixes. Is it OK? Do you like it? :)

waehlisch:
Looks good to me. Other opinions?

@shaddeus, in any case, you are behind the master. Can you also squash both commits into one commit, or start a new PR.

shaddeus:
The Travis CI build failed. Did I make some error?

shaddeus:
I will make a new PR.

waehlisch:
Solved in #44.

Issue #37, opened 2015-11-03
Using two or more groups may cause a double locking of mutex (deadlock)


shaddeus:
Hi!

## Scenario
1. Let have two groups each with one socket and pass these into the RTR MGR
2. Both of two RTR cache server endpoints are down at first and both of two groups go to ```RTR_MGR_ERROR```
3. One of RTR cache server goes up and the group goes to ```RTR_MGR_ESTABLISHED```
4. When a group goes into ```RTR_MGR_ESTABLISHED``` in function ```rtr_mgr_cb()```, immediately is call ```rtr_stop()``` on all sockets in the all others groups that are not ```RTR_MGR_CLOSED```
5. ```rtr_stop()``` invokes ```rtr_change_socket_state()``` and it invokes ```rtr_mgr_cb()``` again -> double locking of ```config->mutex``` (deadlock)

## Backtrace
```gdb
#0  0x00007ffff79c9cfc in __lll_lock_wait () from /usr/lib/libpthread.so.0
#1  0x00007ffff79c3b35 in pthread_mutex_lock () from /usr/lib/libpthread.so.0
#2  0x00007ffff740259a in rtr_mgr_cb (sock=0x6b0f90, state=RTR_SHUTDOWN, data=0x6b1090) at ./rtrlib/rtr_mgr.c:88
#3  0x00007ffff7408044 in rtr_change_socket_state (rtr_socket=0x6b0f90, new_state=RTR_SHUTDOWN)
    at ./rtrlib/rtr/packets.c:189
#4  0x00007ffff7407e83 in rtr_stop (rtr_socket=0x6b0f90) at ./rtrlib/rtr/rtr.c:206
#5  0x00007ffff7402a03 in rtr_mgr_cb (sock=0x6b0e20, state=RTR_ESTABLISHED, data=0x6b1090)
    at ./rtrlib/rtr_mgr.c:139
#6  0x00007ffff7408044 in rtr_change_socket_state (rtr_socket=0x6b0e20, new_state=RTR_ESTABLISHED)
    at ./rtrlib/rtr/packets.c:189
#7  0x00007ffff7407bdc in rtr_fsm_start (rtr_socket=0x6b0e20) at ./rtrlib/rtr/rtr.c:142
#8  0x00007ffff79c14a4 in start_thread () from /usr/lib/libpthread.so.0
#9  0x00007ffff76ff13d in clone () from /usr/lib/libc.so.6
```

### Solution A
[...]

Issue #40, opened 2015-11-03
Go back to the better-preferred group without waiting for actual established group goes to troubles


shaddeus:
## Scenario

1. Let have two groups: Better Group (more preferred) and Worse Group (less preferred). Both include only one socket (cache server).
2. Better Group's cache server is down and Worse Group's cache server is up
3. RTR Manager start
4. RTR Manager synchronizes with Worse Group and **not close Better Group**
5. Better Group's cache server goes up
6. RTR Manager **detects that there is a better option**, so synchronizes with Better Group and close the Worse Group  

## Proposal changes at getting RTR_ESTABLISHED 
- Closing only worse-preferred groups
- If the previous state was RTR_MGR_ERROR then check only better-preferred groups for closed or in error

What do you think about the principle?

PR #41, opened 2015-11-05
Fixes some bugs and improve RTR Manager status callback behavior

shaddeus:
This fixes handling the status callbacks while using more than one ```rtr_mgr_group```. It should call status callbacks in the same manner like using only one ```rtr_mgr_group```.

(Hot)Fixes issue https://github.com/rtrlib/rtrlib/issues/37: deadlock caused by mutex double locking. But I am not sure if the solution is proper.

It adds feature https://github.com/rtrlib/rtrlib/issues/40: Go back to the better-preferred group without waiting for actual established group goes to troubles.

### DIFF
--- a/rtrlib/rtr/rtr.c
+++ b/rtrlib/rtr/rtr.c
@@ -77,6 +77,9 @@ void rtr_init(struct rtr_socket *rtr_socket, struct tr_socket *tr, struct pfx_ta
 
 int rtr_start(struct rtr_socket *rtr_socket)
 {
+    if(rtr_socket->thread_id)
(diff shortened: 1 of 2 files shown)


shaddeus:
The feature #40 brings two bugs:
1. The non-closed error state groups open all groups.
2. Repeated thread creating: It starts threads with already opened rtr_socket.
I will fix it, sorry.

shaddeus:
Hi! I fixed both of the issues.

shaddeus:
Wow, thank you, @reuteran, for accepting the pull request.

PR #43, opened 2015-12-03
doc: update failure handling details in RTR connection manager

waehlisch:
+ Retry Interval
+ Switch back to more preferred group


waehlisch:
@shaddeus, can you please check if the update reflects your changes?

shaddeus:
I agree with this change. The doc is correct.

PR #44, opened 2015-12-04
Prefix all in lib/ with lrtr_

shaddeus:
I am not sure if it is really an improvement.


waehlisch:
@shaddeus, what do you mean, you don't know if this is an improvement?

shaddeus:
@waehlisch, I meant with that only I like more the code without the prefix :)

waehlisch:
Agree, but we need to solve the name clash ;). I think it's the best solution.

Issue #55, opened 2016-04-06
Reduce warnings


waehlisch:
Reduce number of warnings when compiling.

smlng:
Looking at Travis [output](https://travis-ci.org/rtrlib/rtrlib/builds/122530842) for the master branch, I would say we can close this issue for now: @waehlisch, @fho?!

PR #56, opened 2016-04-07
reduce warnings

smlng:
This pull request addresses issue #55 

It fixes all warnings on OSX with clang, but on Linux with GCC some _strict aliasing_ warnings remain.

### DIFF
--- a/rtrlib/rtr/packets.c
+++ b/rtrlib/rtr/packets.c
@@ -369,6 +369,19 @@ static bool rtr_pdu_check_size (const struct pdu_header *pdu) {
     //TODO: Check the error msg 0 termination
 
     break;
+  case SERIAL_QUERY:
+    if (sizeof(struct pdu_serial_query) == pdu->len)
+      retval = true;
+    break;
+  case RESET_QUERY:
+    if (sizeof(struct pdu_reset_query) == pdu->len)
+      retval = true;
+    break;
+  case RESERVED:
+  default:
+    RTR_DBG1("PDU type is unknown or reserved!");
+    retval = false;
+    break;
   }
 
 #ifndef NDEBUG
@@ -704,8 +717,8 @@ static int rtr_undo_update_spki_table(struct rtr_socket *rtr_socket, void *pdu)
 static int rtr_store_prefix_pdu(struct rtr_socket *rtr_socket, const void *pdu, const unsigned int pdu_size, void **ary,
                                 unsigned int *ind, unsigned int *size)
 {
-    const enum pdu_type pdu_type = rtr_get_pdu_type(pdu);
-    assert(pdu_type  == IPV4_PREFIX || pdu_type == IPV6_PREFIX);
+    const enum pdu_type type = rtr_get_pdu_type(pdu);
+    assert(type  == IPV4_PREFIX || type == IPV6_PREFIX);
     if ((*ind) >= *size) {
         *size += TEMPORARY_PDU_STORE_INCREMENT_VALUE;
         void *tmp = realloc(*ary, *size * pdu_size);
@@ -720,9 +733,9 @@ static int rtr_store_prefix_pdu(struct rtr_socket *rtr_socket, const void *pdu, 
         }
         *ary = tmp;
     }
-    if (pdu_type == IPV4_PREFIX)
+    if (type == IPV4_PREFIX)
         memcpy((struct pdu_ipv4 *) *ary + *ind, pdu, pdu_size);
-    else if (pdu_type == IPV6_PREFIX)
+    else if (type == IPV6_PREFIX)
         memcpy((struct pdu_ipv6 *) *ary + *ind, pdu, pdu_size);
     (*ind)++;
     return RTR_SUCCESS;
@@ -731,8 +744,7 @@ static int rtr_store_prefix_pdu(struct rtr_socket *rtr_socket, const void *pdu, 
 static int rtr_store_router_key_pdu(struct rtr_socket *rtr_socket, const void *pdu, const unsigned int pdu_size, void **ary,
                                     unsigned int *ind, unsigned int *size)
 {
-    const enum pdu_type pdu_type = rtr_get_pdu_type(pdu);
-    assert(pdu_type == ROUTER_KEY);
+    assert(rtr_get_pdu_type(pdu) == ROUTER_KEY);
 
     if ((*ind) >= *size) {
         *size += TEMPORARY_PDU_STORE_INCREMENT_VALUE;
@@ -872,7 +884,7 @@ int rtr_sync_receive_and_store_pdus(struct rtr_socket *rtr_socket){
             retval = RTR_ERROR;
             goto cleanup;
         }
-        
+
         type = rtr_get_pdu_type(pdu);
         if (type == IPV4_PREFIX) {
             if (rtr_store_prefix_pdu(rtr_socket, pdu, sizeof(*ipv4_pdus), (void **) &ipv4_pdus, &ipv4_pdus_nindex, &ipv4_pdus_size) == RTR_ERROR){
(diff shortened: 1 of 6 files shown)


smlng:
These _strict aliasing_ warnings are not easily to be fixed, as they refer to operations on packet buffers using uint8_t and uint32_t assignments and conversions. See [rtrlib/rtr/packets.c]() line 210 ff

smlng:
Could someone (@waehlisch, @fho ) review this PR and merge if okay?! I tested on OSX and Debian 8, as mentioned it fixes all but these _strict aliasing_ warnings?

rtrlib/rtr/packets.c
```diff
@@ -369,6 +369,11 @@ static bool rtr_pdu_check_size (const struct pdu_header *pdu) {
     //TODO: Check the error msg 0 termination
 
     break;
+  case SERIAL_QUERY:
+  case RESET_QUERY:
+  case RESERVED:
+  default:
```
fho: The code change removes the warning messages but the checks are still missing for those PDU types.

We should either:
- Implement the missing checks for the PDUs
- keep the warnings, so that we are reminded that something is missing and will add them in the future

rtrlib/rtr_mgr.h
```diff
@@ -73,7 +73,7 @@ typedef void (*rtr_mgr_status_fp)(const struct rtr_mgr_group *, enum rtr_mgr_sta
 /**
  * Type alias \c pfx_for_each_fp for function signature: \code (*)(const struct pfx_record *pfx_record, void *data) \endcode
  */
-typedef void (*pfx_for_each_fp)(const struct pfx_record *pfx_record, void *data);
+//typedef void (*pfx_for_each_fp)(const struct pfx_record *pfx_record, void *data);
```
fho: If the typedef is really not used it should be removed. I think there is no reason to keep commented out code in times of revision control systems :-)

tools/rtrclient.c
```diff
@@ -106,7 +106,7 @@ static void update_spki(struct spki_table* s __attribute__((unused)), const stru
 
 int main(int argc, char** argv)
 {
-    enum mode_t { TCP, SSH } mode;
+    enum mode_t { NOP, TCP, SSH } mode = NOP;
```
fho: It probably fixes a use before initialization warning?
I think it would be better to add an else statement in line 130 and terminate the client here if argv[1] contains a different String than "tcp" or "ssh"

rtrlib/rtr_mgr.h
smlng: Its a duplicate, already defined in [rtrlib/pfx/pfx.h ](https://github.com/rtrlib/rtrlib/blob/c97e0043df737254e67dd23da8264a3e9f363ccf/rtrlib/pfx/pfx.h#L88). However, you're right should be removed completely.

tools/rtrclient.c
smlng: Yes it does fix such warning. Still, I would favor a distinct 'not set' state, e.g. _NOP_ or something similar.

rtrlib/rtr/packets.c
smlng: I would argue that we should have all possible cases covered with the switch statement, even if not implemented yet. But it would be sensible to add some (debug) output that their implementation is missing. On the other hand: is it hard to implement those checks correctly?

If I get this right, they should be implemented anyway - so lets get this right now?

rtrlib/rtr/packets.c
```diff
@@ -731,8 +736,9 @@ static int rtr_store_prefix_pdu(struct rtr_socket *rtr_socket, const void *pdu,
 static int rtr_store_router_key_pdu(struct rtr_socket *rtr_socket, const void *pdu, const unsigned int pdu_size, void **ary,
                                     unsigned int *ind, unsigned int *size)
 {
-    const enum pdu_type pdu_type = rtr_get_pdu_type(pdu);
-    assert(pdu_type == ROUTER_KEY);
+    const enum pdu_type type = rtr_get_pdu_type(pdu);
+    (void) type;
+    assert(type == ROUTER_KEY);
```
nbars: What do you think about replacing this three lines with:
```
assert(rtr_get_pdu_type(pdu) == ROUTER_KEY)
```

rtrlib/rtr/packets.c
nbars: The PDUs are not checked because we will never receive them.
The serial query and reset query PDU are only send by the router.

rtrlib/rtr/packets.c
smlng: I thought about this shorter version, too. But I did want to keep the 'style' consistent with all other functions and just get rid of the warning.

rtrlib/spki/hashtable/ht-spkitable.c
```diff
@@ -73,212 +73,213 @@ static void spki_record_to_key_entry(struct spki_record *spki_r,
...
-		rtval = SPKI_SUCCESS;
-	}
-
-out:
-	pthread_rwlock_unlock(&spki_table->lock);
-	return rtval;
+    uint32_t hash;
+    struct key_entry entry;
+    struct key_entry *rmv_elem;
+    int rtval = SPKI_SUCCESS;
```
nbars: Should be SPKI_ERROR

smlng: Yep, I went through it again and SPKI_ERROR make more sense. My first thought was: most compiler init stuff with 0, and 0 here would be SPKI_SUCCESS. I'll correct that, or has anyone objections?

tools/rtrclient.c
smlng: Btw. the client would then terminate on line 173.

rtrlib/spki/hashtable/ht-spkitable.c
nbars: Variables that are allocated on the stack are never initialized by the compiler.
The value of these variables is unknown!

smlng: So its even more important to have it initialized correctly :-)

rtrlib/rtr/packets.c
waehlisch: @nbars "we will never receive them" -- you never know ;), so throw a debug error.

tools/rtrclient.c
waehlisch: Wrt readability it's better to terminate early, I would say. Concur with @fho suggestion, and remove lines 173-176.

rtrlib/rtr/packets.c
waehlisch: So what do we do?

tools/rtrclient.c
smlng: OK so 2:1, I'll back down ;-)

waehlisch: even worse ;). @nbars was also in favor of @fho proposal.

rtrlib/rtr/packets.c
smlng: No error required, I implemented checks for SERIAL and RESET QUERY - even though they may never be received. For RESERVED a debug error message is shown. PR is updated.

rtrlib/rtr/packets.c
fho: Replace all occurences with the better, shorter version from @nbars ?

waehlisch: OK, let's go for the shorter version.

smlng: all other functions require the ```type``` variable, hence it makes no sense to use shorter version there, see for instance [here](https://github.com/rtrlib/rtrlib/blob/master/rtrlib/rtr/packets.c#L809)

smlng:
I applied all comments, PR is updated and squashed. Please merge, so I can update the PR for travis.

Issue #47, opened 2016-03-16
Improve CI and automatic testing


waehlisch:
- [x] Compiling w/ and w/o SSH in Travis
- [x] Integrate outcome of tests/ into Travis
- [x] Add test function in tests/, which validates against live RTR cache (probably not possible with Travis?)

 

For automatic testing of live data we can verify against the ROA configuration for the RIS beacons, see https://www.ripe.net/analyse/internet-measurements/routing-information-service-ris/current-ris-routing-beacons

smlng:
Compiling with SSH in Travis is possible (just install libssh-dev as a dependency), but why bother compiling _without_ SSH? Failures in non-ssh code parts will be found either way.

Each Travis build environment (VM) has network access, so running live validation tests should be possible, too.

waehlisch:
If (for whatever reasons) SSH is not found, RTRlib might successfully compile w/o SSH. The we assume everything is fine, even though SSH compilation has not been checked.

smlng:
All points are addressed (and fullfilled) by PR #58

smlng:
We can close this one, as soon as PR #58 is merged?

PR #58, opened 2016-04-08
travis with libssh and tests

smlng:
Extended travis configuration to build with libssh and run tests afterwards. It also adds a live validation tests. Note: test_pfx_locks is disabled for ```make test``` as it takes very long to complete.

And fixes to comply with cppcheck.


waehlisch:
Nice @smlng but what about my comment in issue #47? What happens if SSH library was not successfully be installed but non-SSH version compiles successfully?

tests/test_live_validation.c
```diff
@@ -0,0 +1,106 @@
+/*
```
waehlisch: Please use the new header style, see for example rtrlib/rtr_mgr.c

tests/test_live_validation.c
```diff
@@ -0,0 +1,106 @@
...
+ * Website: http://rpki.realmv6.org/
+ */
+
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include <unistd.h>
+#include "rtrlib/rtrlib.h"
+
+#define RPKI_CACHE_HOST "141.22.28.222"
```
waehlisch: Use name rpki-validator.realmv6.org.

tests/test_live_validation.c
```diff
@@ -0,0 +1,106 @@
...
+#define RPKI_CACHE_POST "8282"
+
+typedef struct {
+    char *pfx;
+    int len;
+    int asn;
+    int val;
+} test_validity_t;
+
+const test_validity_t queries[] = {
```
waehlisch: Add comment

> Verification is based on ROAs for RIPE RIS Routing Beacons, see https://www.ripe.net/analyse/internet-measurements/routing-information-service-ris/current-ris-routing-beacons

tests/test_live_validation.c
```diff
@@ -0,0 +1,106 @@
...
+    {"93.175.146.0", 24, 12654, BGP_PFXV_STATE_VALID},      // valid
+    {"84.205.83.0",  24, 12654, BGP_PFXV_STATE_NOT_FOUND},  // not found
+    {"93.175.147.0", 24, 12654, BGP_PFXV_STATE_INVALID},    // invalid
+    {NULL, 0, 0, 0}
```
waehlisch: Include IPv6 prefixes.

tests/test_live_validation.c
smlng: Ah good point, will adopt. Btw. there seems to be some inconsistency between LICENSE (MIT) and file headers (referring to (L)GPL? We should fix that, too - but in a separate PR

tests/test_live_validation.c
smlng: While coding I just remembered the IP, didn't know the corresponding URL anymore. Will fix that.

smlng:
I'll adopt comments above and will update PR soon.

smlng:
Travis fails because of cppcheck, which fails because PR #56 needs to be merged first

CMakeLists.txt
```diff
@@ -53,11 +55,12 @@ ADD_SUBDIRECTORY(tests)
 ENABLE_TESTING()
 ADD_TEST(test_pfx tests/test_pfx)
 ADD_TEST(test_lpfst tests/test_lpfst)
-ADD_TEST(test_pfx_locks tests/test_pfx_locks)
+#ADD_TEST(test_pfx_locks tests/test_pfx_locks)
```
waehlisch: Why is this test not executed anymore?

CMakeLists.txt
```diff
@@ -3,10 +3,12 @@ project(rtrlib)
 cmake_minimum_required(VERSION 2.6)
 
 set(CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/modules)
-
+set(NOSSH false)
```
waehlisch: I'm confused, this does not address my last concern in #47. Suppose that for some reason libssh is not found. Then RTRlib will not be compiled with SSH support. Hence, "cmake -D CMAKE_BUILD_TYPE=NoSSH ." as well as "cmake -D CMAKE_BUILD_TYPE=Release ." will lead to the same outcome. Right?

smlng: On the contrary, I think it does address all points (and concerns) in #47. Further, it is very unlikely, that ```libssh``` is not found _after_ Travis was told to install it explicitly ;-)

smlng: btw. Travis fails not because of my changes, its ```cppcheck``` which fails consistently. So maybe we should disabled it for now in Travis?

fho:
I'll look into it and try to fix the issues that cppcheck reports instead.

CMakeLists.txt
smlng: It takes a lot of time to complete, I ran ```test_pfx_locks``` on my notebook and it took 92 minutes to complete. IMHO such a long, intensive does not qualify to be included in Travis CI testing!

waehlisch:
@smlng hm, it's seems we have a misunderstanding ;). If we assume that installing libssh will not fail (silently), why do we need NoSSH, and more importantly why do we need to run compilation twice. As you wrote before, SSH includes non-SSH.

Anyhow, in the context of testing, arguments such as "very unlikely" are not helpful ;). With our tests we try to cover the unlikely cases ;).

How do we proceed? Do we assume that travis will throw an error in case libssh cannot be installed?

CMakeLists.txt
waehlisch: Agree. However, this significant change should be part of the commit message ;).

smlng:
@waehlisch, lets clear this up a bit:

1. installing _libssh_ will not fail, we use official repos. What could happen is a version mismatch, but for now dependencies of _rtrlib_ are met.
2. Compiling with and without _libssh_ make sense to me (now) - I know, I was against it in the first place ;-) For instance if those parts using _libssh_ have bugs the _release_ build will fail but the _nossh_ build will succeed. 
3.) We can optimize the Travis config by running unit tests only once, and not with and without _libssh_.

fho:
Cppcheck runs now successfully on branch smlng-pr_travis_fix_cppcheck (https://github.com/rtrlib/rtrlib/tree/smlng-pr_travis_fix_cppcheck).

I had to supress a few cppcheck errors that were false or didn't make sense for us.

fho:
cmake reports:

> 0.06s$ cmake -D CMAKE_BUILD_TYPE=Release .
> CMake Warning (dev) in CMakeLists.txt:
>   A logical block opening on the line
>     /home/travis/build/rtrlib/rtrlib/CMakeLists.txt:28 (if)
>   closes on the line
>     /home/travis/build/rtrlib/rtrlib/CMakeLists.txt:37 (endif)
>   with mis-matching arguments.
> This warning is for project developers.  Use -Wno-dev to suppress it.

If the problem was introduced in the branch it should be fixed

smlng:
OK applied all commits to fix cppcheck, thx @fho. FYI: though cppcheck in Travis succeeds, on OS X it still complains about:
```
Checking rtrlib/rtr/packets.c: __cplusplus...
assignBoolToPointer:rtrlib/spki/hashtable/tommyds-1.8/tommyarrayblk.h(92):(error) Boolean value assigned to pointer.
[...]
Checking rtrlib/transport/tcp/tcp_transport.c...
nullPointer:rtrlib/transport/tcp/tcp_transport.c(175):(warning) Possible null pointer dereference: sock - otherwise it is redundant to check it against null.
[...]
unusedFunction:rtrlib/transport/transport.c(39):(style) The function 'tr_ident' is never used.
```

Fixed the issue in CMakeList.txt, too.

However, clear to merge into master now?

waehlisch:
ACK from my side. @fho?

waehlisch:
One minor: Can you squash d510a56. We usually don't include merge of branches as separate commits.

smlng:
@waehlisch: is squashed, as suggested.

CMakeLists.txt
```diff
@@ -23,7 +25,7 @@ set(RTRLIB_SRC rtrlib/rtr_mgr.c rtrlib/lib/utils.c rtrlib/lib/ip.c rtrlib/lib/ip
 set(RTRLIB_LINK ${RT_LIB} ${CMAKE_THREAD_LIBS_INIT})
 
 find_package(LibSSH 0.6.0)
-if(LIBSSH_FOUND)
+if(LIBSSH_FOUND AND NOT NOSSH)
```
fho: searching for the SSH lib could also be disabled by the parameter in cmake/modules/FindLibSSH.cmake

tests/test_live_validation.c
```diff
@@ -0,0 +1,114 @@
+/*
+ * This file is part of RTRlib.
```
fho: I suggest that all new code has to comply with the kernel coding style. Otherwise we have to reformat it again.
All new files should be verified with checkpatch.pl and added to https://github.com/rtrlib/rtrlib/blob/lpfst_cleanup/scripts/check-coding-style.sh to be checked.

tests/test_live_validation.c
```diff
@@ -0,0 +1,114 @@
...
+        } else {
+            printf("%s/%d    \tOK\n", q.pfx, q.len);
+        }
+        q = queries[++i];
+    }
+
+    rtr_mgr_stop(conf);
+    rtr_mgr_free(conf);
+    free(groups[0].sockets);
+
```
fho: tr_free(&tr_tcp); missing?

smlng: I'm not sure, but I have to admit I borrowed many parts from [tools/cli-validator.c](), if its really missing it should be added there as well. However, its defined/initialized as ```struct tr_socket tr_tcp;``` so _free_ is not required, or is it? @fho you are more of an expert here?!

tests/test_live_validation.c
smlng: Will do as soon as PR #59 is merged.

CMakeLists.txt
smlng: @fho: how would I do that? I didn't get it to work :-(

tests/test_live_validation.c
fho: I think it's missing in all tools/ binaries, but im also not 100% sure. :-)

CMakeLists.txt
fho: I thought  adding there the if condition should work, but sounds like you tried it already and it doesn't?

smlng: Nope, didn't tried that - i thought you mean some existing parameter in FindLibSSH.cmake. However, good point and better to have the SSH related stuff in one place. I moved the condition there, its still working. Thx @fho ...

smlng:
Updated PR to latest master including PR #59, tests/test_live_validation.c is kernel coding style compliant. Ready to merge?!

scripts/check-coding-style.sh
```diff
@@ -1,9 +1,15 @@
...
-
+READLINK=$(which greadlink)
+[ -z "$READLINK" ] && {
+	READLINK=$(which readlink)
+}
+[ -z "$1" ] && {
+	echo "USAGE: $0 FILE"
+	exit 1
+}
+CHECKSOURCE=$($READLINK -f "$1")
```
fho: My idea of the script was, that it checks automatically all files that should be checked and can be 1:1 run from travis.
So you don't need knowledge about which files were already converted.
With the change it now takes the files to be checked as commandline parameter?

scripts/cppcheck.sh
```diff
@@ -1,5 +1,9 @@
 #!/bin/bash
 
+<<<<<<< HEAD
 cppcheck --error-exitcode=1 --enable=all --inline-suppr \
 	--template='{id}:{file}({line}):({severity}) {message}' \
 	-i rtrlib/spki/hashtable/tommyds-1.8/ tools/ rtrlib/
+=======
```
fho: The commit seems broken, it should be removed or fixed.

scripts/check-coding-style.sh
```diff
@@ -1,9 +1,15 @@
 #!/bin/bash
 
-script_dir=$(dirname "$(readlink -f "$0")")
-
+READLINK=$(which greadlink)
```
fho: What is greadlink?

smlng: The default ```readlink``` on Linux behaves differently from the one on non-Linux OS such as OS X or Solaris. The latter have _BSD_ ```readlink``` as default while Linux' ```readlink``` is _GNU_ ```readlink```. So on OS X or Solaris you have to explicitly use ```greadlink``` to get the same (_GNU_ variant) behavior as on Linux.

scripts/check-coding-style.sh
smlng: I thought of the script as an helper to run the actual perl code-checker with the right parameters. That way, the script can check any file passed as an argument. 

As our goal is to have all files kernel coding style compliant, we can run something like 
```
find . -name "*tommyds-1.8*" -prune -name "*.c" -o -name "*.h" -exec ./scripts/check-coding-style.sh {} \;
```
in the _rtrlib_ repo-root - and add this command to the travis config.

scripts/cppcheck.sh
smlng: Fixed!

scripts/check-coding-style.sh
fho: Currently only 2 files comply, so we can only check those and fail the build if an error is in those. 
Whenever another file is converted or added it can be added to the list. 
When all files were converted, we could use your find snippet to check all.

smlng: Minor update, the latest count is 3 files, complying to kernel coding style ;-) However, I changed to behavior of the script, either pass a single file to be checked as a parameter OR provide a list of files in ```scripts/files.txt```. This is also add to travis config now.

Hope you (@fho) like that better?!

smlng:
Asking again: good to merge now?

scripts/check-coding-style.sh
```diff
@@ -1,9 +1,32 @@
 #!/bin/bash
 
-script_dir=$(dirname "$(readlink -f "$0")")
+# ---HELP---
+# to check kernel coding style of file(s):
+# a) either pass a filename as cmdline parameter
+# b) create file 'scripts/files.txt' with filenames
+# c) to find all .c and .h files, run in repo root:
+# 	find rtrlib tests tools -name "*tommyds-1.8*" -prune \
+#		-name "*.c" -o -name "*.h" -o -name "*.c"'
```
waehlisch: Why "-name "*.c" twice?

waehlisch:
Almost. Can you please rename  scripts/files.txt to  scripts/check-coding-files.txt, Thanks.

PR #59, opened 2016-04-09
Lpfst cleanup

fho:
I started with cleaning up the code and converting it to kernel coding style.
Please review and merge the branch if there are no objections.


rtrlib/pfx/lpfst/lpfst.c
```diff
@@ -8,151 +8,247 @@
...
+			goto err;
+
+		if (lpfst_get_children(root_node->rchild, array, len) == -1)
+			goto err;
+	}
+
+	return 0;
+
+err:
+	free(*array);
```
smlng: I get a (very) bad feeling, most times I see a function freeing mem it hasn't (m)alloced itself. Wouldn't it be more _safe_ to free ```array``` in the calling function, where it is (re)allocated? This, would also allow us to get rid of the ```goto``` here, which is not necessary, too.

waehlisch:
Many thanks @fho! Looks good to me. @nbars? Regarding Sebastian's comment, I might tend to agree.

rtrlib/pfx/lpfst/lpfst.c
fho: If realloc fails in append_node_to_array(), the memory was not allocated in the same function call.
Therefore it's not the responsibility of append_node_to_array()  to free it.
The caller passed a PTR to a data area to append_node_to_array() and has to decide was should happen if the append_node_to_array() fails. The caller could also decide to handle the error somehow e.g. by sleeping 1ms and try it again.

smlng: OK I think I understand - still I prefer to get around of ```goto``` and freeing foreign memory.

fho: > OK I think I understand - still I prefer to get around of goto and freeing foreign memory.
Could you suggest a different implementation? Maybe I simply don't see the better variant. :-)
I don't agree with the term foreign memory here. The function that frees the memory, allocates the memory in a subsequent function call.
Freeing it in append_node_to_array() would be wrong from my perspective because it wouldn't have been allocated in the same call.

> But maybe reviewing necessity of all goto in the code is something for different PR.
I would suggest the opposite, restructure the code to use goto error handlers.
Goto error handlers are the most clean and readable way to do complicate cleanups sequences.
Without them, you have to use if-statements after the errors and have the same cleanup sequence over and over again. The nesting makes it hard to understand and it's prone to errors because all the duplicated cleanup sequences have to be changes instead of only one

smlng: At first glance I thought that the calling function of ```lpfst_get_children``` will provide an array with pre-allocated memory, and in that case I thought it would be wrong to ```free(*array)```. But as I looked more closely, I understood that a calling function will initially use ```array = NULL``` and ```len = 0```. So that ```lpfst_get_children``` will recursively allocate memory with each child found, right?!

However, due the recursion ```free(*array)``` will be called n-times, with n depending on recursive depth. This is allowed as ```free(NULL)``` will do nothing, but it is still an unnecessary function call. 

Further, this function fails if ```realloc``` fails, which is some kind of memory error (e.g. out-of-mem) - maybe add error/debug message here?

fho: > So that lpfst_get_children will recursively allocate memory with each child found, right?!
right :-)

> However, due the recursion free(*array) will be called n-times, with n depending on recursive depth. This is allowed as free(NULL) will do nothing, but it is still an unnecessary function call.
Yes, the only way to not do the free() in that case would be to add an if-condition before. That would be the same than what free() is calling already internally.
The unnecessary function call won't have any noticable impact, maybe even the compiler inlines it. 

> Further, this function fails if realloc fails, which is some kind of memory error (e.g. out-of-mem) - maybe add error/debug message here?
I don't know to do that without too much effort. 
The best way would be to forward meaningful errno values to the caller. That is currently nowhere done :-(.
Also whenever a memory allocation fails, the kernel will log an out of memory message.

smlng: Hey @fho, how about getting rid of the recursion? Something like this:
```
int lpfst_get_children(const struct lpfst_node *root_node,
		       struct lpfst_node ***array, unsigned int *len)
{
	const struct lpfst_node *node = root_node;
	int pos = 0;
	bool err = false;
	while (!err && node) {
		if (node->lchild) {
			*len += 1;
			if (append_node_to_array(array, len, root_node->rchild))
				err = true;
		}
		if (node->rchild) {
			*len += 1;
			if (append_node_to_array(array, len, root_node->rchild))
				err = true;
		}
		if (!err && (pos < *len))
			node = (*array)[pos++];
		else
			node = NULL;
	}
	if (err) {
		free(*array);
		return -1;
	}
	return 0;
}
```

This way we eliminate ```goto err``` and recursive, unnecessary ```free(*array)```.

fho: For me the variant looks much harder to understand because of much more conditions.
Why you want to replace the goto with an if-statement? Imho the goto is much more straight forward.

smlng: @fho: this is kind of a deadlock: I don't _like_ ```goto``` (+recursive function) you do ;-) Maybe we pick up this discussion again later ...

Leaving this aside, I would recommend to merge this PR and move on. The coding-style test-script would be useful for PR #58, btw.

Issue #3, opened 2013-10-17
Use a sane coding style: kernel coding style


fho:
Convert all sources files to use kernel coding style.
https://www.kernel.org/doc/Documentation/CodingStyle

smlng:
Consistent code style would be great, and the kernel coding style is fine. Except for on thing, I would recommend to use typedef for structs and enums. IMHO this would enhance readability and reduce length of statements.

waehlisch:
They have been removed in PR #8. @fho comments?

smlng:
Yes I see - maybe there is no right or wrong here. So can we find rough consensus on this: do we want to apply _strict_ kernel coding style or revert to/opt for _better_ readability?

Further, in [rtrlib/rtr/packets.c](https://github.com/rtrlib/rtrlib/blob/master/rtrlib/rtr/packets.c) there are lots of _enums_ and _structs_ defined which should be moved to the corresponding header file - or is there any viable reasoning to have these in the C-file?

We could/should apply some code style checker such as [indent](http://www.gnu.org/software/indent/).

fho:
Hello Matthias and Sebastian :-)

> Yes I see - maybe there is no right or wrong here. So can we find rough consensus on this: do we want to apply strict kernel coding style or revert to/opt for better readability?

I would go with the full Kernel coding style, without our own extra rules.

The Kernel Coding Style Guide argues regarding better readability when typedefs are used:
> Chapter 5: Typedefs
> 
> Please don't use things like "vps_t".
> It's a _mistake_ to use typedef for structures and pointers. When you see a
> 
> 	vps_t a;
> 
> in the source, what does it mean?
> In contrast, if it says
> 
> 	struct virtual_container *a;
> 
> you can actually tell what "a" is.
> 
> Lots of people think that typedefs "help readability". Not so. They are
> useful only for: [..]
https://www.kernel.org/doc/Documentation/CodingStyle

If the user wants to have typedefs, he could also declare them on his own.

> Further, in rtrlib/rtr/packets.c there are lots of enums and structs defined which should be moved to the corresponding header file - or is there any viable reasoning to have these in the C-file?
They are in the .c file because they have a private scope. They are only used by the code in that file. They are not intended to be used by other code parts.k

> We could/should apply some code style checker such as indent.
Indent is sadly not enough to convert to kernel coding style. 
Because the code is often nested very deeply,  the 80-characters per line limit (especially with 8 spaces per tab) is not met after automatic reformatting.
[...]

Issue #63, opened 2016-04-14
Update file header license statement


smlng:
The header of most source files is outdated and therefore inconsistent with license (MIT here) issued with  rtrlib as a whole. 

Further, it might be helpful to state/name the (lead) authors of code-files in their header as well - that way issues can be addressed, assigned, and discussed more directly.

smlng:
All done, closing!

Issue #48, opened 2016-03-16
Improve documentation in tests/


waehlisch:
Add documentation to
- [x] test_lpfst.c
- [x] test_pfx.c
- [x] test_pfx_locks.c

As an example see test_ht_spkitable.

smlng:
PR #66 should _fix_ this one. Close issue after merge?

smlng:
Done with merge of PR #66.

PR #66, opened 2016-04-15
tests: fix docu and coding style

smlng:
- update documentation of all tests
- enforce kernel coding style on all tests
- add all tests to coding style check

### DIFF
--- a/tests/test_ht_spkitable.c
+++ b/tests/test_ht_spkitable.c
@@ -1,498 +1,551 @@
 /*
  * This file is part of RTRlib.
  *
- * RTRlib is free software; you can redistribute it and/or modify
- * it under the terms of the GNU Lesser General Public License as published by
- * the Free Software Foundation; either version 3 of the License, or (at your
- * option) any later version.
+ * This file is subject to the terms and conditions of the MIT license.
+ * See the file LICENSE in the top level directory for more details.
  *
- * RTRlib is distributed in the hope that it will be useful, but
- * WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
- * or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public
- * License for more details.
- *
- * You should have received a copy of the GNU Lesser General Public License
- * along with RTRlib; see the file COPYING.LESSER.
+ * Website: http://rtrlib.realmv6.org/
  */
+
 #include "rtrlib/spki/hashtable/ht-spkitable.h"
 #include <stdio.h>
 #include <string.h>
 #include <time.h>
 #include <stdlib.h>
 
-static struct spki_record *create_record(int ASN, int ski_offset, int spki_offset, struct rtr_socket *socket);
+/* test_ht_4 */
+#define NUM_TABLE_X		(50)
+#define NUM_TABLE_Y		(50)
+/* test_ht_6 */
+#define NUM_SKIS		(10)
+#define NUM_SKIS_RECORDS	(10)
+/* test_ht_7 */
+#define NUM_OF_SKI		(200)
+#define NUM_OF_RECORDS		(2000)
 
-static bool compare_spki_records(struct spki_record *r1, struct spki_record *r2)
+/**
+ * @brief Compare SPKI records for equality
+ *
+ * @return true if r1 == r2, false otherwise
+  */
+static bool spki_records_are_equal(struct spki_record *r1,
+				   struct spki_record *r2)
 {
-    if(r1->asn != r2->asn) {
-        return false;
-    }
-    if(r1->socket != r2->socket) {
-        return false;
-    }
-    if(memcmp(r1->ski, r2->ski, SKI_SIZE) != 0) {
-        return false;
-    }
-    if(memcmp(r1->spki, r2->spki, SPKI_SIZE) != 0) {
-        return false;
-    }
-    return true;
-}
+	if (r1->asn != r2->asn)
+		return false;
+	if (r1->socket != r2->socket)
+		return false;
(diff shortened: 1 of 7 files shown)


smlng:
This PR fixes issues #3, #48, and #63 for all tests.

waehlisch:
Thanks @smlng. Can you be a bit more specific in the documentation? For example, "Compare SPKI records" is very short (do you compare equal values, different values etc.?).

smlng:
Update and extend documentation of tests, please review.

waehlisch:
I'm still not very happy ;). For example, 

```
/*
+ * @brief lpfst_test
+ * Run tests to validate a longest prefix first search tree
+ */
```

The additional comment does not give many more insights. What is the test about (instead of testing the lpfst, which is already encoded in the function name)?

Maybe it is a general question: For me the comments should explain the test briefly but in sufficient details such that someone understands what happen without reading the code. What is the input? What is the expected output? However, you can also argue that the code is self-explanatory.

What do we want? @fho @nbars comments?

smlng:
@waehlisch, I elaborated a bit more on the test docu. While it's good to have a short/rough description what each tests, I don't think we want to run doxygen on the test code, or do we?!

waehlisch:
@fho @nbars what do you prefer?

fho:
I agree with @smlng. It's unlikely that someone would read the Doxygen documentation for the tests.
Best would be if the function signature is self-explanatory and we could go without any extra descriptions

smlng:
@fho: please clarify: are the extended docu/comments of tests in this PR sufficient? If so, we should be clear to merge?!

tests/test_ht_spkitable.c
```diff
@@ -1,391 +1,450 @@
...
+static struct spki_record *create_record(int ASN,
+					 int ski_offset,
+					 int spki_offset,
+					 struct rtr_socket *socket);
+/**
+ * @brief Compare SPKI records for equality
+ *
+ * @return true if r1 == r2, false otherwise
+  */
+static bool compare_spki_records(struct spki_record *r1,
```
fho: If the function is renamed to bool spki_records_are_equal(), the function signature would contain the same information than the current description, therefore the description can be omitted.

tests/test_ht_spkitable.c
```diff
@@ -1,391 +1,450 @@
...
+		return false;
+	if (memcmp(r1->spki, r2->spki, SPKI_SIZE) != 0)
+		return false;
+
+	return true;
 }
 
-static struct spki_record *create_record(int ASN, int ski_offset, int spki_offset, struct rtr_socket *socket)
+/**
+ * @brief Create a SPKI record
```
fho: The description doesn't provide any additional value compared to the informations you get out of the function signature.

fho:
I created a few comments in the commit few.

Documentation is always hard to maintain and becomes quickly outdated.
Therefore we should only provide comments if we are not able to make the code self-explanatory.
Having documentation that doesn't add informatinal value is only a burden, makes the project harder to maintain.

tests/test_ht_spkitable.c
```diff
@@ -1,391 +1,450 @@
...
+ *
+ * @return new SPKI record
```
fho: Adds no additionals information, same is provided by the function signature

tests/test_ht_spkitable.c
```diff
@@ -1,391 +1,450 @@
...
+	return record;
 }
 
 /**
- * @brief test_ht_1
- * Test if the spki_table_src_remove function is working correctly by adding
- * spki_record associated with different rtr_socket. Then call spki_table_src_remove
- * with one of the sockets as argument and validate that all records associated with different
- * sockets are still there.
+ * @brief Helper shortcut, to assert insert of table entry
```
fho: Could be helpful to description what does it assert exacty, when will the assert fail?

tests/test_ht_spkitable.c
```diff
@@ -1,391 +1,450 @@
...
-    spki_table_free(&table);
-    free(result);
-    free(socket_one);
-    free(socket_two);
+/* ----- TESTS FUNCTIONS ----- */
 
-    printf("test_h1_1() complete\n");
+/**
+ * @brief test_ht_1
+ * Test if the spki_table_src_remove function is working correctly by adding
```
fho: helpful description, makes it easier to find out what the function is doing

tests/test_pfx.c
```diff
@@ -30,324 +18,404 @@
...
+	}
+
+	/* cleanup: free table */
+	pfx_table_free(&pfxt);
+	printf("mass_test successful\n");
 }
 
-int main()
+/**
+ * @brief Test basic ops on prefix table
```
fho: Which ops are basic?

fho: the brief line can be replaced with the following line.
The first line doesn't provide any information

smlng:
Thanks @fho for reviewing. I included most of your suggestion and updated the PR. 

Even though some comments are redundant with signature of the function they describe; I would rather opt for consistency in documentation, i.e., at least a brief description for each function.

smlng:
No further comments for about a week; @fho, @waehlisch: any objections to merging this one?

fho:
I'm fine with it.
In general it would be good to make smaller commits for each logicial change and explain the commit message why it was done:
e.g.:

- commit for reworking the documentation (commit message could contain why it was decided to also add small descriptions for functions that don't contain additional informations)
- commit for adapting the coding style (should even done per module (tests, lib, pfx, scripts, etc)

It makes it easier to review changes, to understand the reason behind it and allows later to revert  specific changes if needed.

Issue #62, opened 2016-04-11
Checkout coverity for rtrlib code analysis


fho:
Evaluate if https://scan.coverity.com/github is usable to analyze the rtrlib and detect errors. If it is, implement it in the travis build chain

PR #68, opened 2016-04-20
add code coverage

smlng:
This PR addresses issue #62 and integrates http://codecov.io into travis, see for example: https://codecov.io/github/smlng/rtrlib/rtrlib?ref=9d52582ef95a1eb491c475c95cbbf005b2a3e028

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,10 @@ script:
     - make
     - make test
     - make clean
-    - cmake -D CMAKE_BUILD_TYPE=Release .
+    - cmake -D CMAKE_BUILD_TYPE=Release -DENABLE_COVERAGE=On .
     - make
     - make test
+    - make gcov
+
+after_success:
+  - bash <(curl -s https://codecov.io/bash)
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -14,6 +14,10 @@ script:
     - make
     - make test
     - make clean
-    - cmake -D CMAKE_BUILD_TYPE=Release .
+    - cmake -D CMAKE_BUILD_TYPE=Release -DENABLE_COVERAGE=On -ENABLE_COVERAGE_ALL=on .
```
waehlisch: Does `-ENABLE_COVERAGE_ALL=on` not need a `-D`?

smlng:
I followed configuration for cmake from [here](https://github.com/RWTH-ELP/CMake-codecov). The flag seems to be unused (or not required) anyway, so I removed it from the travis config.

waehlisch:
OK. Can you squash the commits? Rest looks fine. 

Btw, I don't know how useful this tool really is but we should give it a try.

smlng:
Squash done! I think it could be quite helpful as it shows which code parts are not covered by any of our tests. So we might be able to improve that ...

Issue #73, opened 2016-05-03
Debian package requires libssh


waehlisch:
It seems that the Debian package requires libssh. We should be more relaxed. If a libssh package is available, we should compile with SSH. If there is no ssh package, we should compile w/o SSH support.

smlng:
It's not a bug, it's hardly a feature. IMHO, RTRlib depending on `libssh` is not a big deal, as the `apt` tools will install any dependencies automatically. However, with PR #74 applied its possible to create deb-packages of RTRlib not relying on `libssh`.

smlng:
Closing this one, is solved with #91 in place.

PR #74, opened 2016-05-04
debian: build deb w/o ssh dependency

smlng:
- (optional) build deb package without libssh dependency
 - add build profile 'nossh'
 - requires dpkg version >= 1.7.14
 - command: dpkg-buildpackage -Pnossh [-us -uc]
 - fix wrong lines in debian/changelog

This PR fixes issue #73. 

_Note_: the build profile disables SSH (dependency on libssh) during deb package build even if the library is installed!

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -14,7 +14,7 @@ librtr0 (0.3.6-1) stable; urgency=low
   * Improve debugging and status values
   * Update SSH transport to new LibSSH API
 
--- Matthias Waehlisch <waehlisch@ieee.org>  Tue,  5 Apr 2016 12:55:56 -0700
+ -- Matthias Waehlisch <waehlisch@ieee.org>  Tue,  5 Apr 2016 12:55:56 -0700
 
 librtr0 (0.3.5-1) stable; urgency=low
 
@@ -27,7 +27,7 @@ librtr0 (0.3.5-1) stable; urgency=low
   * Add support for OS X (10.10)
   * Add new tool cli-validator
 
--- Matthias Waehlisch <waehlisch@ieee.org>  Thu, 20 Aug 2015 04:05:48 -0700
+ -- Matthias Waehlisch <waehlisch@ieee.org>  Thu, 20 Aug 2015 04:05:48 -0700
 
 librtr0 (0.3.0-1) stable; urgency=low
 
--- a/debian/control
+++ b/debian/control
@@ -2,7 +2,7 @@ Source: librtr0
 Section: libs
 Priority: optional
 Maintainer: Fabian Holler <mail@fholler.de>
-Build-Depends: cmake, dpkg-dev (>= 1.16.1~), debhelper (>= 9), libssh-dev (>= 0.6.0), doxygen
+Build-Depends: cmake, dpkg-dev (>= 1.16.1~), debhelper (>= 9), libssh-dev (>= 0.6.0) <!nossh>, doxygen
 Standards-Version: 3.9.4
 Vcs-Git: git://github.com/rtrlib/rtrlib.git
 Vcs-Browser: https://github.com/rtrlib/rtrlib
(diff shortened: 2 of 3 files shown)


waehlisch:
Two things:

1. Is this compatible with Ubuntu 12.04?
1. Is there no way to compile RTRlib with and w/o SSH support, depending on the availability of libssh?

fho:
2.) Debian packages containing the binary of the application. The packages are build from a central build manager, which builds the package in a chroot and preinstalls all required dependency for the build.
Building libssh depending on the availablity of libssh wouldn't make sense in this scenario. Libssh would never be available if not explcitiy declared as dependency in the rules file.

I also don't see an advantage for having 2 different versions of the debian package. Normaly simply senseful compile options are choosing, depending on e.g. overhead, cost and usefulness for the majority of users.
Having rtrlib with libssh makes sense for most users and the overhead is negligible.

smlng:
We should sort this out a bit, I think I misunderstood something. This PR allows to _manually_ generate Debian-Packages of precompiled RTRlib with and w/o depending on `libssh`.

However, if I (now) understand correctly we want (or at least @waehlisch 😉) 2 versions of RTRlib in the package repos: one with `libssh` and one without; something like `librtr0`, `librtr-dev` and `librtr0-nossh`  - correct? However, as @fho pointed out; I don't see the point to have two version of `librtr`.

Regarding the other points:

1. No, `dpkg` Version in Ubuntu 12.04 LTS is 1.6.1. 
2. as @fho said: the build environment is initialized according to the description and dependencies defined in the `debian/control` file. If we specify `libssh` as a dependency it will be available during package build. And if it is not available on the destined host it will be installed by `apt-get` as a dependency.

smlng:
I took another look into this, and from what I read in related (dpkg) docs it is not possible to build (sub)packages with differing build options (in our case: w/ libssh and w/o libssh) in a single deb-package configuration. The process to create the deb-packages can be roughly summarized as follows:

1. read configuration in `debian/*` files
2. create build environment accordingly
3. build binaries from source as specified in `debian/rules` (this happens only once!)
4. create deb packages as specified in `debian/control`

Hence, a `librtr0-nossh` deb-package requires a separate configuration (`debian_nossh/...`), do we really need/want that? I'm with @fho, as I don't understand why depending on `libssh` is a problem. Its just a minor overhead (in terms of disk space) and any developer using RTRlib is still free to use simple TCP _or_ SSH transport.

fho:
Should we then only merge the changes in debian/changelog and discard the introduction of the build profile?

smlng:
The major problem I see here is that resulting packages w/ or w/o libssh dependency are indistinguishable (by their names). This could lead to unexpected failures, e.g. when someone installs RTRlib w/o SSH support (not knowing), but some tools require RTRlib with SSH. 

And even if we create a separate RTRlib deb-package w/o libssh dependency, all programs that depend on RTRlib would have to specify which version of RTRlib they require _or_ have to specify libssh as an additional dependency. 

IMHO, we should provide a reasonable default deb-package of RTRlib, i.e. with libssh, and any special requirements are either met by the build profile (if we agree to provide it) or build the lib from source w/o libssh.

waehlisch:
Sorry for being silent on this topic! To clarify the motivation. The feature request resulted from collaboration with @mwinter-osr from Quagga. They operate a comprehensive testing infrastructure, where they test Quagga-RTRlib on multiple platforms. One is Ubuntu 12.04, which only supports libssh-dev package 0.5.2 but RTRlib requires 0.6.0. Consequently, building RTRlib package fails because of this and that the RTRlib package requires libssh. So a solution that does not work under Ubuntu 12.04 does not help.

smlng:
Ok, so this is not about having rtrlib w/o ssh support in the apt repos, but supporting (an out-dated, 4 year old version of) Ubuntu, namely 12.04 LTS. For the latter, there are 14.04 LTS and recently 16.04 LTS. 

I see two or three solutions:

1. use backport-repos, there seems to be libssh 0.6.1 in them, see [here](http://www.ubuntuupdates.org/package/core/precise/main/updates/libssh)
2. manually build libssh > 0.6 on Ubuntu 12.04
3. [or manually build rtrlib w/o ssh support on Ubuntu 12.04]

Or can we downgrade rtrlib to libssh 0.5.something?

waehlisch:
OK option 2. Sorry for the noise,

smlng:
Ping?! From the discussion above I assume this one is outdated/unwanted and can be closed?

smlng:
I close this one. It is outdated and with #91 merged soon, rtrlib support Ubuntu 12.04 with libssh 0.5.

PR #79, opened 2016-06-20
packets: save new interval values only after session_id check

shaddeus:
Hi, developers.

I found maybe a small bug. It used to save new received interval values even from End-Of-Data PDUs with bad session_id. A Session ID check should be performed before processing intervals.

### DIFF
--- a/rtrlib/rtr/packets.c
+++ b/rtrlib/rtr/packets.c
@@ -914,21 +914,21 @@ int rtr_sync_receive_and_store_pdus(struct rtr_socket *rtr_socket){
             RTR_DBG1("EOD PDU received.");
             struct pdu_end_of_data_v0 *eod_pdu = (struct pdu_end_of_data_v0 *) pdu;
 
-            if (eod_pdu->ver == RTR_PROTOCOL_VERSION_1) {
-                rtr_socket->expire_interval = ((struct pdu_end_of_data_v1 *) pdu)->expire_interval;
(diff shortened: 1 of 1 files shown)


shaddeus:
I'm thinking about it a little more now. The cache servers are trusted, so an update timing interval values  is all right even from a bad EOD PDU, isn't it?

waehlisch:
@shaddeus regarding your second comment, no, the spec explicitly says "If, at any time after the
protocol version has been negotiated (Section 7), either the router or the cache finds the value of the Session ID is not the same as the other's, the party which detects the mismatch MUST immediately terminate the session with an Error Report PDU with code 0 ("Corrupt Data"), and the router MUST flush all data learned from that cache." 

Currently, the implementation sets the timer values before checking the Session ID, but when a Session ID mismatch occurs, it resets (almost immediately). What you propose is an optimization in the implementation (don't set the timer values if you already know that there is a problem).

So, good catch, we should accept this little optimization.

nbars:
Currently there are no checks implemented to validate if the received interval values are in the ranges specified by the RFC. Should we refuse to set "invalid" interval values?

waehlisch:
@nbars right, we only check in rtr.c.Can you open an issue or submit a PR?

Luckily, it's not a security issue as there is a trusted relation between cache and router.

shaddeus:
@waehlisch fine! Thanks for merging and the explanation of proper behaviour in that case.

PR #83, opened 2016-06-20
codecov: add yml config file

smlng:
codecov.io now support yml configuration file within source repo, this PR copies online settings and adds respective config file.


waehlisch:
Can we name the file .codecov.yml  ?

smlng:
First, I wasn't sure if this allow, but it is. So be it 😉

PR #85, opened 2016-07-01
Add simple linking troubleshooting

ColinBS:
On some systems the path to the rtrlib.so must be passed to gcc using -L,
else the compiler complains about not finding lrtr.

### DIFF
--- a/README
+++ b/README
@@ -75,7 +75,7 @@ occurs, the location of the library can be passed explicitly as a parameter
 
     -L<path_to_librtr.so>
 
-eg.,
+e.g.,
     -L/usr/local/lib64/
 
 
--- a/README
+++ b/README
@@ -66,6 +66,18 @@ pass the following parameter to gcc:
 
     -lrtr
 
+In case an error such as
+
+    -/usr/bin/ld: cannot find -lrtr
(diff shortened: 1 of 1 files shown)


README
```diff
@@ -66,6 +66,18 @@ pass the following parameter to gcc:
 
     -lrtr
 
+In case an error like this
+
+    -/usr/bin/ld: cannot find -lrtr
+    -collect2: error: ld returned 1 exit status
+
+occurs, the location of the library must be passed as a parameter
```
waehlisch: s/must/can/

README
```diff
@@ -66,6 +66,18 @@ pass the following parameter to gcc:
...
+
+    -L<path_to_librtr.so>
+
+eg.
```
waehlisch: e.g.,

README
```diff
@@ -66,6 +66,18 @@ pass the following parameter to gcc:
 
     -lrtr
 
+In case an error like this
```
waehlisch: s/like this/such as/

nbars:
I would suggest to merge these two commits into a single one.
The commit "Add suggestions from pull request" doesn't add any valuable
information's for someone who is reading the commit history.

PR #86, opened 2016-07-07
Proposal for new configuration API of the rtrlib

nbars:
This pull request is intended to serve as a discussion platform, thus please don't merge it!
Please add comments if something is unclear or if you have a better proposal.

Decisions I'am not sure about:
- The ability to change interval values (e.g. timeout) per cache?

A usage example can be found at the bottom of the file.

### DIFF
--- a/rtrlib/rtr/new-api.h
+++ b/rtrlib/rtr/new-api.h
@@ -105,3 +105,30 @@ int rtr_mgr_stop(rtr_mgr_t *mgr);
  * @param mgr The rtr_mgr_t that should be freed.
  */
 void rtr_mgr_free(rtr_mgr_t *mgr);
+
+
+//Example calling code
+rtr_transport_socket_t *tpc = rtr_transport_tcp_socket_create(
+                              "6.6.6.6",
+                              1234,
+                              NULL);
+
+rtr_mgr_group_t *group = rtr_group_add_cache_socket(
+                         NULL,
+                         tcp,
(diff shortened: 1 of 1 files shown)


rtrlib/rtr/new-api.h
```diff
@@ -0,0 +1,136 @@
+//Internal types
+struct rtr_transport_socket;
+struct rtr_mgr_group;
+struct rtr_mgr;
+
+//external types
+typedef struct rtr_transport_socket rtr_transport_socket_t;
+typedef struct rtr_mgr_group rtr_mgr_group_t;
+typedef struct rtr_mgr rtr_mgr_t;
+
```
fho: 1. We decided to use the kernel coding style. The style guide says:

> 5) Typedefs
> -----------
> 
> Please don't use things like ``vps_t``.
> It's a **mistake** to use typedef for structures and pointers. When you see a
> 
> .. code-block:: c
> 
> 
> 	vps_t a;
> 
> in the source, what does it mean?
> In contrast, if it says
> 
> .. code-block:: c
> 
> 	struct virtual_container *a;
> 
> you can actually tell what ``a`` is.
https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/plain/Documentation/process/coding-style.rst

2. The POSIX standard forbids to use _t for variable names:
https://stackoverflow.com/a/12727104/537958

PR #91, opened 2016-09-26
rtrlib: LibSSH 0.5.0 support

ColinBS:
The CMakeLists.txt now checks for SSH 0.5.0 support.
@smlng hinted that the ssh_userauth_publickey_auto function of SSH 0.6.0 is more or less the same as the older 0.5.0 version ssh_userauth_autopubkey.
Still needs testing to confirm same behavior of these functions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,11 @@ cmake_minimum_required(VERSION 2.6)
 set(CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/modules)
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -std=gnu99 -fstack-protector-all")
 if(CMAKE_BUILD_TYPE STREQUAL Debug)
-    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu99 -Wall -Wextra -Wfree-nonheap-object -Wformat-security -Wmissing-prototypes -Wmissing-declarations -Wdeclaration-after-statement -Winit-self -Waggregate-return -Wmissing-format-attribute -Wundef -Wbad-function-cast -Wwrite-strings -Wformat=2")
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu99 -Wall -Wextra -Wformat-security -Wmissing-prototypes -Wmissing-declarations -Wdeclaration-after-statement -Winit-self -Waggregate-return -Wmissing-format-attribute -Wundef -Wbad-function-cast -Wwrite-strings -Wformat=2")
+    execute_process(COMMAND ${CMAKE_C_COMPILER} -dumpversion OUTPUT_VARIABLE GCC_VERSION)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -80,7 +89,7 @@ install(DIRECTORY rtrlib/ DESTINATION include/rtrlib
 
 #pkgconfig file
 if(LIBSSH_FOUND)
-    set (PKG_CONFIG_REQUIRES "libssh >= 0.6.0")
+    set (PKG_CONFIG_REQUIRES "libssh >= 0.5.2")
```
smlng: should `libssh >= 0.5.0` suffice, too? Or wouldn't it be even better?

rtrlib/transport/ssh/ssh_transport.c
```diff
@@ -73,7 +73,12 @@ int tr_ssh_open(void *socket)
         goto error;
     }
 
-    const int rtval = ssh_userauth_publickey_auto(ssh_socket->session, NULL, NULL);
+    const int rtval;
+#ifdef LIBSSH_060
+    rtval = ssh_userauth_publickey_auto(ssh_socket->session, NULL, NULL);
+#elif defined LIBSSH_052
```
smlng: if its not libssh 0.6.x (or above) it has to be 0.5.x, otherwise `cmake` would have failed first. So you can simplify this to `#else` - and may add a short comment something like:
```
#ifdef LIBSSH_060
...
#else // LIBSSH_050
...
#endif
```

waehlisch:
Please always squash merge commits.

doxygen/examples/rtr_mgr.c
```diff
@@ -1,6 +1,38 @@
...
+static void connection_status_callback(const struct rtr_mgr_group *group,
+				       enum rtr_mgr_status status,
+				       const struct rtr_socket *socket,
+				       void *data)
+{
+	if (status == RTR_MGR_ERROR)
+		connection_status = status;
+}
+
+int connection_error(enum rtr_mgr_status status)
```
smlng: this function is not used anywhere, see `cli-validator` where it is used.

tools/cli-validator.c
```diff
@@ -16,6 +16,8 @@
 const int connection_timeout = 20;
 enum rtr_mgr_status connection_status = -1;
 
+int connection_error(enum rtr_mgr_status status);
```
smlng: is this prototype needed?

doxygen/examples/rtr_mgr.c
```diff
@@ -73,14 +106,38 @@ int main(){
     rtr_mgr_start(conf);
 
     //wait till at least one rtr_mgr_group is fully synchronized with the server
-    while(!rtr_mgr_conf_in_sync(conf))
+    while(!rtr_mgr_conf_in_sync(conf)) {
+	if (connection_status)
```
smlng: this will likely fail immediately, because connection_status is init with `-1`, or not?

smlng:
please squash

PR #97, opened 2016-10-28
ci: fail on style issues, check level issues are ignored

mroethke:
checkpatch does not differentiate between the check levels in the return code, as soon as one fails it returns with 1.
I patched it to ignore "check" level issues regarding the return code, they are still printed, to prevent ci fails over stuff that is outside of our control.

### DIFF
--- a/scripts/check-coding-style.sh
+++ b/scripts/check-coding-style.sh
@@ -15,6 +15,7 @@ READLINK=$(which greadlink)
 }
 SCRIPT_DIR=$(dirname "$($READLINK -f "$0")")
 SCRIPT_FILE="$SCRIPT_DIR/check-coding-files.txt"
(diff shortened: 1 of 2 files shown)


fho:
I assume the issue was that the CI failed because checkpatch.pl failed on CHECK:PREFER_KERNEL_TYPES issues.

1. We should not modify checkpatch.pl. The checkpatch.pl script is continously improved in the linux kernel. If we change it, it makes it harder to replace our version with a newer one. We would always have to adapt it to our changes again and again.

2. If checkpatch.pl complains about something that doesn't make sense in our case, the check can be disabled by passing the "--ignore" parameter to checkpatch.pl, e.g.:
`checkpatch --ignore=PREFER_KERNEL_TYPES `

Issue #81, opened 2016-06-20
Missing validation of interval values received from cache server


nbars:
Currently there are no checks implemented to validate if the received interval values are in the ranges specified by the RFC. See #79 and `rtr_sync_receive_and_store_pdus` in `packets.c`

waehlisch:
Fixed in #98.

PR #98, opened 2016-10-28
rtrlib/rtr: add range checks for interval values in packets.c

colinbs:
Closes issue #81
Uses same if-statements used in rtr.c
There still are return values for failed checks needed, suggestions are welcome.

### DIFF
--- a/rtrlib/rtr/packets.c
+++ b/rtrlib/rtr/packets.c
@@ -178,6 +178,12 @@ static int rtr_send_error_pdu_from_host(const struct rtr_socket *rtr_socket,
                                         const char *err_text,
                                         const uint32_t err_text_len);
 
+static int interval_send_error_pdu(struct rtr_socket *rtr_socket,
+                                   void *pdu,
+                                   uint32_t interval,
+                                   uint16_t minimum,
+                                   uint32_t maximum);
+
 static inline enum pdu_type rtr_get_pdu_type(const void *pdu)
 {
     return *((char *) pdu + 1);
@@ -190,6 +196,78 @@ static int rtr_set_last_update(struct rtr_socket *rtr_socket)
         rtr_change_socket_state(rtr_socket, RTR_ERROR_FATAL);
         return RTR_ERROR;
     }
+    return RTR_SUCCESS;
+}
+
+int rtr_check_interval_range(uint32_t interval, uint32_t minimum,
+			     uint32_t maximum)
+{
+    if (interval < minimum) {
+        return BELOW_INTERVAL_RANGE;
+    } else if (interval > maximum) {
+        return ABOVE_INTERVAL_RANGE;
+    }
+
+    return INSIDE_INTERVAL_RANGE;
+}
+
+void apply_interval_value(struct rtr_socket *rtr_socket,
+			  uint32_t interval,
+			  enum interval_type type)
+{
+    if (type == EXPIRATION) {
+        rtr_socket->expire_interval = interval;
+    } else if (type == REFRESH) {
+        rtr_socket->refresh_interval = interval;
+    } else if (type == RETRY) {
+        rtr_socket->retry_interval = interval;
+    }
+}
+
+int rtr_check_interval_option(struct rtr_socket *rtr_socket,
+			      int interval_mode,
+			      uint32_t interval,
+			      enum interval_type type)
+{
+    uint16_t minimum;
+    uint32_t maximum;
+
+    int interv_retval;
+
+    switch (type) {
+    case EXPIRATION:
+        minimum = RTR_EXPIRATION_MIN;
+        maximum = RTR_EXPIRATION_MAX;
+        interv_retval = rtr_check_interval_range(interval, minimum, maximum);
+        break;
+    case REFRESH:
+        minimum = RTR_REFRESH_MIN;
+        maximum = RTR_REFRESH_MAX;
+        interv_retval = rtr_check_interval_range(interval, minimum, maximum);
+        break;
+    case RETRY:
+        minimum = RTR_RETRY_MIN;
+        maximum = RTR_RETRY_MAX;
+        interv_retval = rtr_check_interval_range(interval, minimum, maximum);
+        break;
+    default:
+        RTR_DBG("Invalid interval type: %u.", type);
+        return RTR_ERROR;
+    }
+
+    if (interv_retval == INSIDE_INTERVAL_RANGE ||
+	interval_mode == ACCEPT_ANY) {
+        apply_interval_value(rtr_socket, interval, type);
+    } else if (interval_mode == DEFAULT_MIN_MAX) {
+        if (interv_retval == BELOW_INTERVAL_RANGE)
+            apply_interval_value(rtr_socket, minimum, type);
+        else
+            apply_interval_value(rtr_socket, maximum, type);
+    } else {
+        RTR_DBG("Received expiration value out of range. Was %u. "
+                    "It will be ignored.", interval);
+    }
+
     return RTR_SUCCESS;
 }
 
@@ -797,7 +875,7 @@ static int rtr_store_prefix_pdu(struct rtr_socket *rtr_socket, const void *pdu, 
  * @attention ary is not freed in this case, because it might contain data that is still needed
  */
 static int rtr_store_router_key_pdu(struct rtr_socket *rtr_socket, const void *pdu, const unsigned int pdu_size,
-				    struct pdu_router_key **ary, unsigned int *ind, unsigned int *size)
+                    struct pdu_router_key **ary, unsigned int *ind, unsigned int *size)
 {
     assert(rtr_get_pdu_type(pdu) == ROUTER_KEY);
 
@@ -919,8 +997,8 @@ int rtr_sync_receive_and_store_pdus(struct rtr_socket *rtr_socket){
     unsigned int ipv6_pdus_size = 0;
 
     struct pdu_ipv4 *ipv4_pdus = NULL;
-    unsigned int ipv4_pdus_size = 0; //next free index in ipv4_pdus
-    unsigned int ipv4_pdus_nindex = 0;
+    unsigned int ipv4_pdus_size = 0; 
+    unsigned int ipv4_pdus_nindex = 0; //next free index in ipv4_pdus
 
     struct pdu_router_key *router_key_pdus = NULL;
     unsigned int router_key_pdus_size = 0;
@@ -970,10 +1048,34 @@ int rtr_sync_receive_and_store_pdus(struct rtr_socket *rtr_socket){
                 goto cleanup;
             }
 
-            if (eod_pdu->ver == RTR_PROTOCOL_VERSION_1) {
-                rtr_socket->expire_interval = ((struct pdu_end_of_data_v1 *) pdu)->expire_interval;
-                rtr_socket->refresh_interval = ((struct pdu_end_of_data_v1 *) pdu)->refresh_interval;
-                rtr_socket->retry_interval = ((struct pdu_end_of_data_v1 *) pdu)->retry_interval;
+            if (eod_pdu->ver == RTR_PROTOCOL_VERSION_1
+                    && rtr_socket->iv_mode != IGNORE_ANY) {
+
+                int interv_retval;
+                interv_retval = rtr_check_interval_option(rtr_socket, rtr_socket->iv_mode, ((struct pdu_end_of_data_v1 *) pdu)->expire_interval, EXPIRATION);
+
+                if(interv_retval == RTR_ERROR) {
+                    interval_send_error_pdu(rtr_socket, pdu, ((struct pdu_end_of_data_v1 *) pdu)->expire_interval, RTR_EXPIRATION_MIN, RTR_EXPIRATION_MAX);
+                    retval = RTR_ERROR;
+                    goto cleanup;
+                }
+
+                interv_retval = rtr_check_interval_option(rtr_socket, rtr_socket->iv_mode, ((struct pdu_end_of_data_v1 *) pdu)->refresh_interval, REFRESH);
+
+                if(interv_retval == RTR_ERROR) {
+                    interval_send_error_pdu(rtr_socket, pdu, ((struct pdu_end_of_data_v1 *) pdu)->refresh_interval, RTR_REFRESH_MIN, RTR_REFRESH_MAX);
+                    retval = RTR_ERROR;
+                    goto cleanup;
+                }
+
+                interv_retval = rtr_check_interval_option(rtr_socket, rtr_socket->iv_mode, ((struct pdu_end_of_data_v1 *) pdu)->retry_interval, RETRY);
+
+                if(interv_retval == RTR_ERROR) {
+                    interval_send_error_pdu(rtr_socket, pdu, ((struct pdu_end_of_data_v1 *) pdu)->retry_interval, RTR_RETRY_MIN, RTR_RETRY_MAX);
+                    retval = RTR_ERROR;
+                    goto cleanup;
+                }
+
                 RTR_DBG("New interval values: expire_interval:%u, refresh_interval:%u, retry_interval:%u",
                         rtr_socket->expire_interval, rtr_socket->refresh_interval, rtr_socket->retry_interval);
             }
@@ -1024,10 +1126,10 @@ int rtr_sync_receive_and_store_pdus(struct rtr_socket *rtr_socket){
                         retval = rtr_undo_update_pfx_table(rtr_socket, &(ipv4_pdus[j]));
                     for (unsigned int j = 0; j < ipv6_pdus_nindex && retval == PFX_SUCCESS; j++)
                         retval = rtr_undo_update_pfx_table(rtr_socket, &(ipv6_pdus[j]));
-		     // cppcheck-suppress duplicateExpression
+             // cppcheck-suppress duplicateExpression
                     for (unsigned int j = 0; j < i && (retval == PFX_SUCCESS || retval == SPKI_SUCCESS); j++)
                         retval = rtr_undo_update_spki_table(rtr_socket, &(router_key_pdus[j]));
-		     // cppcheck-suppress duplicateExpression
+             // cppcheck-suppress duplicateExpression
                     if (retval == RTR_ERROR || retval == SPKI_ERROR) {
                         RTR_DBG1("Couldn't undo all update operations from failed data synchronisation: Purging all key entries");
                         spki_table_src_remove(rtr_socket->spki_table, rtr_socket);
@@ -1151,7 +1253,6 @@ int rtr_wait_for_sync(struct rtr_socket *rtr_socket)
     return RTR_ERROR;
 }
 
-
 static int rtr_send_error_pdu(const struct rtr_socket *rtr_socket, const void *erroneous_pdu,
 		       const uint32_t erroneous_pdu_len, const enum pdu_error_type error,
 		       const char *err_text, const uint32_t err_text_len)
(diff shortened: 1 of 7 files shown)


waehlisch:
Please, do **not** merge. There are several open items:

- [ ]  There is no appropriate error handling.
- [ ]  There is no error log message in case that these values overwrite pre-configured values (see rtr.c). Actually, it is unclear to me if overwriting pre-configured values is a good option at all. I contacted Randy and Rob.

PS: Please, don't use verb conjugation (e.g., use "add ..." instead "adds ...") in commity message.

smlng:
@ColinBS, I agree with @waehlisch, before we merge we need _real_ error handling. Further, though haven't had the time for a careful review, I think those values have to be passed through `ntohl` see [here](https://github.com/ColinBS/rtrlib/blob/a77d9a6263e551829cec9bf936fc55601b203611/rtrlib/rtr/packets.c#L235).

waehlisch:
Some more feedback after talking with Rob.

(1) Should the router send an error PDU to the cache server if the cache server sends a timer value exceeding the spec values?
 
He suggests that the "client could just enforce the minimum and maximum values (eg, if the cache says something larger than max, router treats it as if cache had said max).  I suppose one might have a knob to control whether the router does this or instead generates an Error."

Personally, I like the idea but it should be very well documented. The only error PDU that currently might fit is Corrupt Data (fatal).

(2)  If the admin configured explicitly other values compared to the values sent by the cache, can the cache overwrite these values?

There is a clear statement that the router should not harm the cache. (Note that there is a trust relation between cache and router.)

By default, the router should accept the values communicated by the cache. However, in any case, a log message should be fired.

Furthermore, there should be a configuration knob that allows the router to keep the explicitly configured expiration interval.

smlng:
If we want to have the behavior to be configurable this will atleast introduce a new call to the API, we should be careful there. Thus, for now I would recommend to just extend the API by a separate, additional call. If we change/uptdate/optimize the rtrlib API in the future (see #4, #86), this can be put into the config of the rtr_mgr.

ColinBS:
So I guess it should be left as it is for now and come back to this as soon as the API receives changes.

smlng:
@ColinBS: as I said above, an additional call -that is, no modification to existing API calls- should be safe right now. However, it might not be optimal but that can be fixed as soon as we update/cleanup the API.

ColinBS:
Several things were changed/added:
To avoid magic numbers for the min/max values for interval boundaries an enum was added which contains these numbers.
There now is a (temporary) solution on, if desired, forcing boundaries should recieved values be out of bounds. This is still WIP and only for demonstration/discussion purposes.
New functions rtr_check_expiration/refresh/retry_range() check if received values are inside boundaries or not. This makes the code more readable. A new enum interval_range offers return values for these functions.

rtrlib/rtr/rtr.h
```diff
@@ -27,6 +27,16 @@
 #define RTR_DBG(fmt, ...) lrtr_dbg("RTR Socket: " fmt, ## __VA_ARGS__)
 #define RTR_DBG1(a) lrtr_dbg("RTR Socket: " a)
 
+// min and max values for expiration time
+static const uint16_t EXPIRATION_MIN = 600;     // ten minutes
```
smlng: Use prefix `RTR_` for all `#defines`

rtrlib/rtr/rtr.c
```diff
@@ -64,19 +64,19 @@ int rtr_init(struct rtr_socket *rtr_socket,
     if(tr != NULL)
         rtr_socket->tr_socket = tr;
 
-    if(refresh_interval > 86400 || refresh_interval < 1) {
+    if(refresh_interval > REFRESH_MAX || refresh_interval < REFRESH_MIN) {
```
smlng: why not use the (new) ´rtr_check_interval_range` function here and below?

rtrlib/rtr/packets.c
```diff
@@ -168,6 +191,37 @@ static inline enum pdu_type rtr_get_pdu_type(const void *pdu)
     return *((char *) pdu + 1);
 }
 
+static int rtr_check_interval_range(uint32_t interval, const int interval_type)
```
smlng: move this function to `rtrlib/rtr.c` to use it in other code parts as well, see below.

rtrlib/rtr/packets.c
```diff
@@ -19,6 +19,7 @@
 
 #define TEMPORARY_PDU_STORE_INCREMENT_VALUE 100
 #define MAX_SUPPORTED_PDU_TYPE 10
+#define INTERVAL_OPTION -1
```
smlng: this will not work, should be `static int interval_option = SOME_DEFAULT_VALUE`, and there should be a API call to set this parameter.

rtrlib/rtr/packets.c
```diff
@@ -923,10 +977,98 @@ int rtr_sync_receive_and_store_pdus(struct rtr_socket *rtr_socket){
...
-                rtr_socket->expire_interval = ((struct pdu_end_of_data_v1 *) pdu)->expire_interval;
-                rtr_socket->refresh_interval = ((struct pdu_end_of_data_v1 *) pdu)->refresh_interval;
-                rtr_socket->retry_interval = ((struct pdu_end_of_data_v1 *) pdu)->retry_interval;
+            if (eod_pdu->ver == RTR_PROTOCOL_VERSION_1
+                    && INTERVAL_OPTION != IGNORE_ANY) {
+                int interv_retval = rtr_check_interval_range(((struct pdu_end_of_data_v1 *) pdu)->expire_interval, EXPIRATION);
+                if (interv_retval == INSIDE_INTERVAL_RANGE) {
+                    rtr_socket->expire_interval = ((struct pdu_end_of_data_v1 *) pdu)->expire_interval;
+                } else {
+                    switch (INTERVAL_OPTION) {
```
smlng: won't work, see above. Use static variable.

rtrlib/rtr/packets.c
```diff
@@ -923,10 +977,98 @@ int rtr_sync_receive_and_store_pdus(struct rtr_socket *rtr_socket){
...
+                            rtr_socket->refresh_interval = EXPIRATION_MIN;
+                        } else {
+                            rtr_socket->refresh_interval = EXPIRATION_MAX;
+                        }
+                        break;
+                    case ACCEPT_ANY:
+                        rtr_socket->expire_interval = ((struct pdu_end_of_data_v1 *) pdu)->expire_interval;
+                        break;
+                    case IGNORE_ON_FAILURE:
+                        // if interval values are out of bounds, ignore them without error PDU.
```
smlng: add debug output here, too

rtrlib/rtr/packets.c
```diff
@@ -923,10 +977,98 @@ int rtr_sync_receive_and_store_pdus(struct rtr_socket *rtr_socket){
...
+                        goto cleanup;
+                        break;
+                    }
+                }
+
+                interv_retval = rtr_check_interval_range(((struct pdu_end_of_data_v1 *) pdu)->refresh_interval, RETRY);
+                if (interv_retval == INSIDE_INTERVAL_RANGE) {
+                    rtr_socket->retry_interval = ((struct pdu_end_of_data_v1 *) pdu)->retry_interval;
+                } else {
+                    switch (INTERVAL_OPTION) {
```
smlng: Here is a lot of code duplication to handle the different intervals according to the interval_option.

Either write another helper function that just returns the correct value corresponding to `interval_option` and `interval_type` or expand/rewrite `rtr_check_interval_range` to return the value instead of just `-1`, `0`, or `1`,

rtrlib/rtr/packets.c
```diff
@@ -47,6 +48,28 @@ enum pdu_type {
     ERROR = 10
 };
 
+// These modes let the user configure how
+// received intervals should be handled.
+enum interval_mode {
```
smlng: this enum might be needed for a future config API call, so might be good to move it into a header file.
Further, please add doxygen conform documentation for this enum then.

rtrlib/rtr/packets.c
```diff
@@ -20,6 +20,8 @@
 #define TEMPORARY_PDU_STORE_INCREMENT_VALUE 100
 #define MAX_SUPPORTED_PDU_TYPE 10
 
+static int interval_option = DEFAULT_MIN_MAX;
```
smlng: This is only visible here, but you also define it in `rtr.c`, hence it should be globally defined (and set) somewhere

rtrlib/rtr/rtr.c
```diff
@@ -17,6 +17,8 @@
 #include "rtrlib/rtr/rtr.h"
 #include "rtrlib/lib/utils.h"
 
+static int interval_option = DEFAULT_MIN_MAX;
```
smlng: see comment above

rtrlib/rtr/rtr.c
```diff
@@ -52,6 +54,99 @@ static int install_sig_handler()
...
+        break;
+    case REFRESH:
+        minimum = RTR_REFRESH_MIN;
+        maximum = RTR_REFRESH_MAX;
+        break;
+    case RETRY:
+        minimum = RTR_RETRY_MIN;
+        maximum = RTR_RETRY_MAX;
+        break;
+    default:
```
smlng: add error message and return here?

rtrlib/rtr/rtr.c
```diff
@@ -52,6 +54,99 @@ static int install_sig_handler()
...
+        break;
+    }
+
+    int interv_retval = rtr_check_interval_range(interval, type);
+
+    if (interv_retval == INSIDE_INTERVAL_RANGE || interval_mode == ACCEPT_ANY) {
```
smlng: I would move the `switch() case` here, and check for `INSIDE_INTERVAL_RANGE` where required

rtrlib/rtr/rtr.c
```diff
@@ -64,22 +159,28 @@ int rtr_init(struct rtr_socket *rtr_socket,
     if(tr != NULL)
         rtr_socket->tr_socket = tr;
 
-    if(refresh_interval > 86400 || refresh_interval < 1) {
-        return RTR_INVALID_PARAM;
-    } else {
+    int interv_retval = rtr_check_interval_range(refresh_interval, REFRESH);
+
+    if(interv_retval == INSIDE_INTERVAL_RANGE) {
```
smlng: you could summarize the check into one like:
```
if (rtr_check_interval_range(refresh_interval, REFRESH) != INSIDE_INTERVAL_RANGE || 
    rtr_check_interval_range(refresh_interval, EXPIRATION) != INSIDE_INTERVAL_RANGE || 
    rtr_check_interval_range(refresh_interval, RETRY) != INSIDE_INTERVAL_RANGE) {
    /* optional: print some error message? */
    return RTR_INVALID_PARAM;
}
/* otherwise not error, hence set all as given */
```

rtrlib/rtr/rtr.h
```diff
@@ -39,6 +48,35 @@ enum rtr_rtvals {
...
+    REFRESH = 1,
+    RETRY = 2
+};
+
+/**
+ * @brief These modes let the user configure how received intervals should be handled.
+ */
+enum interval_mode {
+    /** Ignore appliance of interval values at all. */
+    IGNORE_ANY = 0,
```
smlng: add comments for doxygen like this:
```
IGNORE_ANY = 0,     /*< Ignore appliance of interval values at all. */
```

fho:
merged

rtrlib/rtr/rtr.h
ColinBS: Should I adjust the comments for the rtr_socket_state enum as well, while I'm at it? Or are these not relevant for Doxygen?

rtrlib/rtr/rtr.c
```diff
@@ -52,6 +53,127 @@ static int install_sig_handler()
...
+        break;
+    case IGNORE_ON_FAILURE:
+        interval_option = option;
+        txt = "IGNORE_ON_FAILURE";
+        break;
+    default:
+        RTR_DBG1("Invalid interval mode. Mode remains unchanged.");
+        return;
+    }
+    RTR_DBG("Interval mode set to %s", txt);
```
smlng: This functions complexity is only related to this debug output, otherwise it should be just:
```
void set_interval_mode(int option)
{
    switch(option) {
    case IGNORE_ANY:
    case ACCEPT_ANY:
    case DEFAULT_MIN_MAX:
    case IGNORE_ON_FAILURE:
        interval_option = option;
        break;
    default:
         RTR_DBG1("Invalid interval mode. Mode remains unchanged.");
    }
}
```

rtrlib/rtr/rtr.c
```diff
@@ -52,6 +53,127 @@ static int install_sig_handler()
...
+        RTR_DBG("Interval mode set to: %u", get_interval_mode());
+    } else {
+        RTR_DBG1("Invalid interval mode. Mode remains unchanged.");
+    }
+     */
+}
+
+int rtr_check_interval_range(uint32_t interval, int16_t minimum, int32_t maximum)
+{
+    if(interval < minimum) {
```
smlng: add space `if (` and braces are not discouraged by kernel coding style

rtrlib/rtr/rtr.c
```diff
@@ -52,6 +53,127 @@ static int install_sig_handler()
...
+                                     enum interval_type type) {
+    uint16_t minimum;
+    uint32_t maximum;
+
+    int interv_retval;
+
+    switch (type) {
+    case EXPIRATION:
+        minimum = RTR_EXPIRATION_MIN;
+        maximum = RTR_EXPIRATION_MAX;
```
smlng: minimum, and maximum can be initialized before switch?

rtrlib/rtr/rtr.h
```diff
@@ -39,6 +48,31 @@ enum rtr_rtvals {
...
+    REFRESH = 1,
+    RETRY = 2
+};
+
+/**
+ * @brief These modes let the user configure how received intervals should be handled.
+ */
+enum interval_mode {
+    IGNORE_ANY = 0,         /*< Ignore appliance of interval values at all. */
+
```
smlng: remove empty lines, here and below

rtrlib/rtr/rtr.c
```diff
@@ -52,6 +53,127 @@ static int install_sig_handler()
...
+        interv_retval = rtr_check_interval_range(interval, minimum, maximum);
+        break;
+    default:
+        RTR_DBG("Invalid interval type: %u.", type);
+        return RTR_ERROR;
+    }
+
+    if (interv_retval == INSIDE_INTERVAL_RANGE || interval_mode == ACCEPT_ANY) {
+        apply_interval_value(rtr_socket, interval, type);
+    } else {
```
smlng: IMHO this would be readable without `switch () case` rewrite for instance like:

```
if (interv_retval == INSIDE_INTERVAL_RANGE || interval_mode == ACCEPT_ANY) {
    apply_interval_value(rtr_socket, interval, type);
}
if else (interval_mode == DEFAULT_MIN_MAX) {
    if (interv_retval == BELOW_INTERVAL_RANGE)
        apply_interval_value(rtr_socket, minimum, type);
    else
        apply_interval_value(rtr_socket, maximum, type);
}
else {
    RTR_DBG();
}

File: tests/unittests/test_packets_static.c [hunk 20]
```diff
@@ -271,6 +271,151 @@ static void test_rtr_pdu_check_size(void **state)
...
+			 pdu_eod->refresh_interval);
+
+	apply_interval_value(rtr_socket, pdu_eod->retry_interval, RETRY);
+	assert_int_equal(rtr_socket->retry_interval, pdu_eod->retry_interval);
+
+	apply_interval_value(rtr_socket, pdu_eod->expire_interval, EXPIRATION);
+	assert_int_equal(rtr_socket->expire_interval, pdu_eod->expire_interval);
+
+	/* test checks that determine if value is inside range */
+	retval = 2;
```
smlng: why is `retval` set here, or below, when it is replaced by function return value afterwards?

rtrlib/rtr/rtr.c
ColinBS: To prevent confusion on my side: should minimum and maximum be initialized before the switch statement like this?
`uint16_t minimum = 0;`
`uint32_t maximum = 0;`

smlng: ah sorry, forget about it thought min/max are always the same, but they are not :/ my bad.

rtrlib/rtr/rtr.h
```diff
@@ -39,6 +48,31 @@ enum rtr_rtvals {
...
 };
 
+enum interval_range {
+    BELOW_INTERVAL_RANGE = -1,
+    INSIDE_INTERVAL_RANGE = 0,
+    ABOVE_INTERVAL_RANGE = 1
+};
+
+enum interval_type {
+    EXPIRATION = 0,
```
smlng: explicite values are not needed here, enums start with `0` by default and increment by `1`, too.

rtrlib/rtr/rtr.h
smlng: and same here, enums start with 0 and increment by 1 as a default.

rtrlib/rtr/rtr.h
```diff
@@ -170,5 +204,48 @@ void rtr_stop(struct rtr_socket *rtr_socket);
...
+/**
+ * @brief Checks if a given interval is inside or outside of the predefined ranges.
+ * @param[in] interval Interval value received from a data PDU.
+ * @param[in] minimum Lower bounds of the range.
+ * @param[in] maximum Upper bounds of the range.
+ * @return BELOW_INTERVAL_RANGE If the given interval is below the specified range.
+ * @return ABOVE_INTERVAL_RANGE If the given interval is above the specified range.
+ * @return INSIDE_INTERVAL_RANGE If the given interval is inside the specified range.
+ */
+int rtr_check_interval_range(uint32_t interval, int16_t minimum, int32_t maximum);
```
smlng: I think its a bit strange that min and max have different types, `int16_t` and `int32_t` specifically, which could be unsigned by the way, right?!

colinbs: Yeah, it looks kind of strange. I used `int16_t` since the minimum values for the intervals fit into it. Not sure if this is even necessary.
Anyway, I will change it to `int32_t` and use `unsigned`.

smlng: the intervals are stored as `uint32_t` in the socket struct anyway - so makes sense to use unsigned

smlng:
please rebase

smlng:
@colinbs ping?

tests/unittests/test_packets_static.c
```diff
@@ -353,6 +353,144 @@ static void test_rtr_send_error_pdu(void **state)
...
+					   pdu_eod->refresh_interval, REFRESH);
+	assert_int_equal(retval, RTR_SUCCESS);
+	assert_int_equal(rtr_socket->refresh_interval, RTR_REFRESH_MAX + 1);
+}
+
+static void test_set_interval_option(void **state)
+{
+	UNUSED(state);
+
+	struct rtr_socket *rtr_socket = malloc(1024);
```
mroethke: dito

tests/unittests/test_packets_static.c
```diff
@@ -353,6 +353,144 @@ static void test_rtr_send_error_pdu(void **state)
...
+	assert_int_equal(rtr_socket.refresh_interval, RTR_REFRESH_MAX + 1);
+}
+
+static void test_set_interval_option(void **state)
+{
+	UNUSED(state);
+
+	struct rtr_socket rtr_socket;
+
+	set_interval_mode(&rtr_socket, IGNORE_ANY);
```
smlng: you need to add an include for `rtr_mgr.h`, Travis reports compiler warning:

```
/rtrlib/rtrlib/tests/unittests/test_packets_static.c:478:2: warning: implicit declaration of function ‘set_interval_mode’ [-Wimplicit-function-declaration]
  set_interval_mode(&rtr_socket, IGNORE_ANY);
  ^
```

smlng:
please squash the 2 unittest commits into one

smlng:
@mroethke if you agree, please merge (without squash)

Issue #99, opened 2016-11-17
Incorrect validation results


reuteran:
For prefix 1.9.23.0/24 there are two ROAs:
```
ASN: 4788
Prefix: 1.9.0.0/16
Max length: 24
```

```
ASN: 65120
Prefix: 1.9.23.0/24
Max length: 24
```

I used `'pfx_table_validate_r(..)' ` to validate this:
```
ASN: 4788
Prefix: 1.9.23.0/24
```

And the result is `BGP_PFXV_STATE_INVALID`, but it should be `BGP_PFXV_STATE_VALID`.

I've confirmed this using the cli-validator:

```
rtrlib/tools $ ./cli-validator rpki-validator.realmv6.org 8282
Arguments required: IP Mask ASN
1.9.23.0 24 4788 //Input
1.9.23.0 24 4788|65120 1.9.23.0 24 24|2 //Output, 2 means INVALID
1.9.23.0 24 65120 //Input
1.9.23.0 24 65120|65120 1.9.23.0 24 24|0 //Output, 0 means VALID
```

I checked the ROAs at http://rpki-validator.realmv6.org:9090/roas (search for 1.9.23.0/24)

It looks like the validation code is only checking the ROA with the most specific prefix match, and if that ASN doesn't match it says INVALID.

waehlisch:
You are right. That's an embarrassing bug. Can someone check the code and fix this ASAP?

@reuteran did you check that the /16 ROA is part of the pfx_table?

Actually, if you are right, the current code searches for the most specific covering (not matching) ROA.

reuteran:
@waehlisch 

> did you check that the /16 ROA is part of the pfx_table?

Not directly, since gdb won't let me step into it for some reason (compiled with debug symbols).
But I queried for:
```
asn: 4788
prefix: 1.9.0.0/16
```
which returns:
```
1.9.0.0 16 4788|4788 1.9.0.0 16 24|0
```
The section after the first '|' means that there is indeed an pfx_table entry for 1.9.0.0/16-24 for AS4788.

waehlisch:
OK. Still embarrassing. Can you fix this?

reuteran:
OK, but it might take a bit because I am unfamiliar with the pfx_table. Hopefully I'll have it by tomorrow

smlng:
I added a test in #101 that should reflect the described case, however the test passes successful not showing the same problem.

waehlisch:
[...]

PR #101, opened 2016-11-19
tests: extend pfx test

smlng:
This extends `test_pfx.c` with the case discussed in #99. However, it does not show the described false validation result.


smlng:
@waehlisch and @reuteran please verify that the added test resembles the issue in #99.

smlng:
requires #103 to succeed

smlng:
suggested extra tests are part of #112, so closing

PR #102, opened 2016-11-21
hashtable: remove assembly files because they are not used

mroethke:
It seems the assembly code is completely unnecessary, all test run without it and none of our make files has any reference to it.
It is not even "real" assembly code but gnu assembler which is just the, platform independent, default backend of the gcc.

As far as I can see the only thing it is doing right now is tricking github into thinking that rtrlib is an assembly project due to the huge amount of assembly code.
According to sloccount about 34k of asm vs 7300 of c code.


smlng:
@fho IIRC you introduced this hashtable into the code, are you okay with removing the assembly files, or is there any license issue or anything? From what I see (and Travis confirms) they are not needed.

smlng:
No further comments or objections by other, so ACK and merge.

PR #103, opened 2016-11-22
Add 'trie' data structure as alternative to lpfst

### LINKED ISSUE #99: Incorrect validation results

reuteran:
This fixes #99 by using a trie instead of a lpfst. Using a trie, the lookup during validation will find all covering prefixes and check their ROAs.

In this patch the trie is the default data structure, to fall back on lpfst the user has to set "RTRLIB_USE_LPFST" in the main CMakeList.txt

### DIFF
--- a/rtrlib/pfx/trie/trie-pfx.c
+++ b/rtrlib/pfx/trie/trie-pfx.c
@@ -366,9 +366,9 @@ int pfx_table_validate(struct pfx_table *pfx_table, const uint32_t asn, const st
 
 int pfx_table_src_remove(struct pfx_table *pfx_table, const struct rtr_socket *socket)
 {
+    pthread_rwlock_wrlock(&(pfx_table->lock));
     for(unsigned int i = 0; i < 2; i++) {
         struct trie_node **root = (i == 0 ? &(pfx_table->ipv4) : &(pfx_table->ipv6));
-        pthread_rwlock_wrlock(&(pfx_table->lock));
         if(*root != NULL) {
             int rtval = pfx_table_remove_id(pfx_table, root, *root, socket, 0);
             if(rtval == PFX_ERROR) {
@@ -376,8 +376,8 @@ int pfx_table_src_remove(struct pfx_table *pfx_table, const struct rtr_socket *s
                 return PFX_ERROR;
             }
         }
-        pthread_rwlock_unlock(&pfx_table->lock);
     }
+    pthread_rwlock_unlock(&pfx_table->lock);
     return PFX_SUCCESS;
 }
 
--- a/rtrlib/pfx/trie/trie-pfx.c
+++ b/rtrlib/pfx/trie/trie-pfx.c
@@ -1,0 +1,472 @@
+/*
+ * This file is part of RTRlib.
+ *
+ * This file is subject to the terms and conditions of the MIT license.
+ * See the file LICENSE in the top level directory for more details.
+ *
+ * Website: http://rtrlib.realmv6.org/
+ */
+
+#include "rtrlib/pfx/trie/trie-pfx.h"
+#include <assert.h>
+#include <pthread.h>
+#include <stdlib.h>
+#include <stdbool.h>
+
+struct data_elem {
+    uint32_t asn;
+    uint8_t max_len;
+    const struct rtr_socket *socket;
+};
+
+struct node_data {
+    unsigned int len;
+    struct data_elem *ary;
+};
+
+static struct trie_node *pfx_table_get_root(const struct pfx_table *pfx_table, const enum lrtr_ip_version ver);
+static int pfx_table_del_elem(struct node_data *data, const unsigned int index);
+static int pfx_table_create_node(struct trie_node **node, const struct pfx_record *record);
+static int pfx_table_append_elem(struct node_data *data, const struct pfx_record *record);
+static struct data_elem *pfx_table_find_elem(const struct node_data *data, const struct pfx_record *record, unsigned int *index);
+static bool pfx_table_elem_matches(struct node_data *data, const uint32_t asn, const uint8_t prefix_len);
+static void pfx_table_notify_clients(struct pfx_table *pfx_table, const struct pfx_record *record, const bool added);
+static int pfx_table_remove_id(struct pfx_table *pfx_table, struct trie_node **root, struct trie_node *node, const struct rtr_socket *socket, const unsigned int level);
+static int pfx_table_node2pfx_record(struct trie_node *node, struct pfx_record records[], const unsigned int ary_len);
+static void pfx_table_free_reason(struct pfx_record **reason, unsigned int *reason_len);
+
+
+void pfx_table_notify_clients(struct pfx_table *pfx_table, const struct pfx_record *record, const bool added)
+{
+    if(pfx_table->update_fp != NULL)
+        pfx_table->update_fp(pfx_table, *record, added);
+}
+
+void pfx_table_init(struct pfx_table *pfx_table, pfx_update_fp update_fp)
+{
+    pfx_table->ipv4 = NULL;
+    pfx_table->ipv6 = NULL;
+    pfx_table->update_fp = update_fp;
+    pthread_rwlock_init(&(pfx_table->lock), NULL);
+}
+
+void pfx_table_free(struct pfx_table *pfx_table)
+{
+    for(int i = 0; i < 2; i++) {
+        struct trie_node *root = (i == 0 ? pfx_table->ipv4 : pfx_table->ipv6);
+        if(root != NULL) {
+            struct trie_node *rm_node;
+
+            pthread_rwlock_wrlock(&(pfx_table->lock));
+            do {
+                struct node_data *data = (struct node_data *) (root->data);
+                for(unsigned int i = 0; i < data->len; i++) {
+                    struct pfx_record record = { data->ary[i].asn, (root->prefix), root->len, data->ary[i].max_len, data->ary[i].socket};
+                    pfx_table_notify_clients(pfx_table, &record, false);
+                }
+                rm_node = (trie_remove(root, &(root->prefix), root->len, 0));
+                assert(rm_node != NULL);
+                free(((struct node_data *) rm_node->data)->ary);
+                free(rm_node->data);
+                free(rm_node);
+            } while(rm_node != root);
+            if(i == 0)
+                pfx_table->ipv4 = NULL;
+            else
+                pfx_table->ipv6 = NULL;
+
+            pthread_rwlock_unlock(&(pfx_table->lock));
+        }
+    }
+    pthread_rwlock_destroy(&(pfx_table->lock));
+}
+
+int pfx_table_append_elem(struct node_data *data, const struct pfx_record *record)
+{
+    data->len++;
+    data->ary = realloc(data->ary, sizeof(struct data_elem) * data->len);
+    if(data->ary == NULL)
+        return PFX_ERROR;
+    data->ary[data->len - 1].asn = record->asn;
+    data->ary[data->len - 1].max_len = record->max_len;
+    data->ary[data->len - 1].socket = record->socket;
+    return PFX_SUCCESS;
+}
+
+int pfx_table_create_node(struct trie_node **node, const struct pfx_record *record)
+{
+    *node = malloc(sizeof(struct trie_node));
+    if(*node == NULL)
+        return PFX_ERROR;
+    (*node)->prefix = record->prefix;
+    (*node)->len = record->min_len;
+    (*node)->lchild = NULL;
+    (*node)->rchild = NULL;
+    (*node)->parent = NULL;
+
+    (*node)->data = malloc(sizeof(struct node_data));
+    if((*node)->data == NULL)
+        return PFX_ERROR;
+    ((struct node_data *) (*node)->data)->len = 0;
+    ((struct node_data *) (*node)->data)->ary = NULL;
+
+    return pfx_table_append_elem(((struct node_data *) (*node)->data), record);
+}
+
+struct data_elem *pfx_table_find_elem(const struct node_data *data, const struct pfx_record *record, unsigned int *index)
+{
+    for(unsigned int i = 0; i < data->len; i++) {
+        if(data->ary[i].asn == record->asn && data->ary[i].max_len == record->max_len && data->ary[i].socket == record->socket) {
+            if(index != NULL)
+                *index = i;
+            return &(data->ary[i]);
+        }
+    }
+    return NULL;
+}
+
+//returns pfx_table->ipv4 if record version is LRTR_IPV4 else pfx_table->ipv6
+inline struct trie_node *pfx_table_get_root(const struct pfx_table *pfx_table, const enum lrtr_ip_version ver)
+{
+    return ver == LRTR_IPV4 ? pfx_table->ipv4 : pfx_table->ipv6;
+}
+
+int pfx_table_del_elem(struct node_data *data, const unsigned int index)
+{
+    //if index is not the last elem in the list, move all other elems backwards in the array
+    if(index != data->len - 1) {
+        for(unsigned int i = index; i < data->len - 1; i++) {
+            data->ary[i] = data->ary[i+1];
+        }
+    }
+    data->len--;
+    data->ary = realloc(data->ary, sizeof(struct data_elem) * data->len);
+    if(data->len != 0 && data->ary == NULL)
+        return PFX_ERROR;
+    else if(data->len == 0)
+        data->ary = NULL;
+
+    return PFX_SUCCESS;
+}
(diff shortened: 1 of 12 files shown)


waehlisch:
@reuteran, did you push test_trie.c? Doesn't seem so.

reuteran:
That was it, my bad.

waehlisch:
@ColinBS, can you please write a script (chain) that compares the validation outcome of the RIPE Validator (http://rpki-validator.realmv6.org:9090/export) with the validation outcome of the RTRlib? Please, identify those prefixes that have different validation outcome. As input for the IP prefixes, you can use a recent RIB dump of route-views2 (http://archive.routeviews.org/bgpdata/). Should not take much time.

smlng:
@waehlisch wouldn't it be more sensible to compare the validation results of the `cli-validator` tool using _RTRlib with lpfst_ vs. _RTRlib with trie_? 

Because `rpki-validator.realmv6.org` is basically the input/cache for `cli-validator`, so there should be no difference.

waehlisch:
No. We want to verify that the Trie is implemeted correctly, i.e., finds the correct ROAs for a BGP Update.
(RIPE Validator also verifies BGP Updates, you have to use the API.)

smlng:
Ah, using the API not the exported files, I misunderstood first. 

Nevertheless, comparing to RTRlib with lpfst make sense as well, that way we know the error margin of the original/old version?!

rtrlib/pfx/trie/trie-pfx.c
```diff
@@ -0,0 +1,472 @@
...
+            pthread_rwlock_unlock(&(pfx_table->lock));
+        }
+    }
+    pthread_rwlock_destroy(&(pfx_table->lock));
+}
+
+int pfx_table_append_elem(struct node_data *data, const struct pfx_record *record)
+{
+    data->len++;
+    data->ary = realloc(data->ary, sizeof(struct data_elem) * data->len);
```
nbars: Memory leaks at data->ary if realloc returns NULL

rtrlib/pfx/trie/trie-pfx.c
```diff
@@ -0,0 +1,472 @@
...
+            else
+                pfx_table->ipv6 = NULL;
+
+            pthread_rwlock_unlock(&(pfx_table->lock));
+        }
+    }
+    pthread_rwlock_destroy(&(pfx_table->lock));
+}
+
+int pfx_table_append_elem(struct node_data *data, const struct pfx_record *record)
```
nbars: I think this function should be R/W locked

rtrlib/pfx/trie/trie-pfx.c
```diff
@@ -0,0 +1,472 @@
...
+        return PFX_ERROR;
+    (*node)->prefix = record->prefix;
+    (*node)->len = record->min_len;
+    (*node)->lchild = NULL;
+    (*node)->rchild = NULL;
+    (*node)->parent = NULL;
+
+    (*node)->data = malloc(sizeof(struct node_data));
+    if((*node)->data == NULL)
+        return PFX_ERROR;
```
nbars: Leak at *node

rtrlib/pfx/trie/trie-pfx.c
```diff
@@ -0,0 +1,472 @@
...
+{
+    //if index is not the last elem in the list, move all other elems backwards in the array
+    if(index != data->len - 1) {
+        for(unsigned int i = index; i < data->len - 1; i++) {
+            data->ary[i] = data->ary[i+1];
+        }
+    }
+    data->len--;
+    data->ary = realloc(data->ary, sizeof(struct data_elem) * data->len);
+    if(data->len != 0 && data->ary == NULL)
```
nbars: Leak at data->ary if realloc fails

rtrlib/pfx/trie/trie-pfx.c
```diff
@@ -0,0 +1,472 @@
...
+
+int pfx_table_validate(struct pfx_table *pfx_table, const uint32_t asn, const struct lrtr_ip_addr *prefix, const uint8_t prefix_len, enum pfxv_state *result)
+{
+    return pfx_table_validate_r(pfx_table, NULL, NULL, asn, prefix, prefix_len, result);
+}
+
+int pfx_table_src_remove(struct pfx_table *pfx_table, const struct rtr_socket *socket)
+{
+    for(unsigned int i = 0; i < 2; i++) {
+        struct trie_node **root = (i == 0 ? &(pfx_table->ipv4) : &(pfx_table->ipv6));
```
nbars: The pointer ```pfx_table->ipvX``` will point to an invalid memory address if someone removes the last node form ```pfx_table->ipvX``` and ```pfx_table->ipvX``` gets freed because it was the last node.

So, you should get the R/W lock before getting the reference.

rtrlib/pfx/trie/trie-pfx.c
nbars: Could be possible that you're calling this function only from locked code sections.

reuteran: Yes, this function is only called from within trie-pfx.c and only by functions which have the lock already.

reuteran:
@nbars Updated some of the changes. Please review, I will then change it in the lpfst file as well.

fho:
Could you please shortly explain in the commit message that introduces the datastructure how it works?

reuteran:
@fho @smlng Since I already pushed these commits, is there a way to safely amend their commit messages without using force push? Or should I squash them together to one commit and write the explanation there? Does the "Squash and merge" button in github allow me to write a commit message?

fho:
Because it's your private working branch and no other branches are based on it, you can freely rebase and force push as much as you want. :-)
I would prefer to have multiple commits for each logical change, e.g. add test for X, add trie datastructure, fix memory leak in X, fix possible invalid memory X at Y, etc

waehlisch:
Somehow I agree. However, splitting d77af5565022509e6594fc6a7a24edba67fa72ed now might introduce too much effort. You can also argue that introducing a new data structure requires tests etc. @fho maybe a compromise: @reuteran extends the commit message, enumerating all substeps that are part of the commit.

fho:
I don't think it would be so much work to split it into separate commits. The logic changes weren't big, most renaming can be done with search&replace.
I see it as very important to have proper commits. It makes it much easier to follow the changes for others and also for ourselfs when we forgot the details in 2 weeks.
It often requires a little bit additional work to rebase, fixup and split commits but it's worth it.

Also we added the file https://github.com/rtrlib/rtrlib/blob/master/CONTRIBUTING to the repository that links commit guidelines that we wanted to be followed. So we also should follow them. :-)

If the rework is considered as too much work. I could also finish my branch of the same idea (https://github.com/rtrlib/rtrlib/commits/issue/99).

Regarding the tests, we have a lot of tests that are also testing the new datastructure in test_pfx.c and test_trie.c. Also my commit https://github.com/rtrlib/rtrlib/commit/1261e3bca55cb3da44bd45eaa8a6e6b8b048e663 can be cherry-picked, it contains adapted tests  for the trie datastructure (functions still have to be renamed in the test file).

reuteran:
Ok, I will separate the commits into smaller chunks later this week.

fho:
awesome :-)

fho:
What do you think about removing the lpfst code completly instead of introducing the RTRLIB_USE_LPFST define to activate/deactivate the lpfst?

The LPFST doesn't work correctly for the validation. So I don't know a reason to keep the code activable.
In the unlikely case that we need the lpfst code again in the future, we can recover it from an older commit.

waehlisch:
@fho I was also thinking about removing. However, it may be helpful to keep it for performance-data comparisons.

fho:
I would recommend to always remove unused code.
Keeping unused code increases the complexity of the software. It has to be updated and tested when the API changes. It also simply increases the amount of code that has to be read to understand the software.
Especially when somebody new get into the project it's often not immediately clear if the code has any legitimate purpose and time is wasted with it.

Often those kind of code will never be used again. Everybody only wonders in the future why it is still there, everybody is unsure if it can be removed or if there is a good reason to keep it.

If we would need in the future for performance test, we could either use an older version of rtrlib that uses the lpfst or recover the lpfst code from an older commit and integrate it in that version

reuteran:
@fho 
> For which purpose should we keep the old lpfst_lookup() when it's more efficient to use trie_lookup_first_match() for validation?

I don't think there is a purpose, I just included it in the beginning because I wanted to keep the API the same between the two data structures. I've actually removed the lookup and renamed the lookup_first_match to lookup, fixed the doc too. 

@fho  @waehlisch @smlng About lpfst removal:
Maybe we can have a separate branch for performance testing, and remove lpfst from the master branch. Or just remove it from the master branch and if we want to performance test we can checkout a previous commit as fabian said. In any case I think removal of lpfst should happen separate from the addition of the trie.

Is there anything that still needs to be done for this PR before it can be merged?

fho:
@reuteran 

1.) Sorry, I overlooked your "Remove trie_lookup(..) and rename trie_lookup_first_match(..) to trie_lookup(..)" commit somehow. :-)


2.) If we decide that we want to remove the RTRLIB_USE_LPFST define, it should be removed in this branch.
Otherwise we merge a change that introduces new code that is only there to be cleaned up in the future.
Also I don't like the idea that we have a master version that enables the user to build rtrlib with 2 datastructures, where one of it is fundamentally broken.
The functionality that we offer should work.
If we would would like to support multiple pfx datastructures. We have to ensure that always both are working. This includes changing the travis build script to make builds with both variants. 
I think removing the lpfst datastructures and the define before the merge is the easiest and quickest way to have something consistent in master that works.



3.) The coding style issues should be fixed:
```
> check coding style of rtrlib/pfx/trie/trie.h ...
rtrlib/pfx/trie/trie.h:40: CHECK:PARENTHESIS_ALIGNMENT: Alignment should match open parenthesis
rtrlib/pfx/trie/trie.h:56: CHECK:PARENTHESIS_ALIGNMENT: Alignment should match open parenthesis
rtrlib/pfx/trie/trie.h:74: CHECK:PARENTHESIS_ALIGNMENT: Alignment should match open parenthesis
rtrlib/pfx/trie/trie.h:88: CHECK:PARENTHESIS_ALIGNMENT: Alignment should match open parenthesis
rtrlib/pfx/trie/trie.h:101: CHECK:PARENTHESIS_ALIGNMENT: Alignment should match open parenthesis
> check coding style of rtrlib/pfx/trie/trie.c ...
rtrlib/pfx/trie/trie.c:59: CHECK:PARENTHESIS_ALIGNMENT: Alignment should match open parenthesis
rtrlib/pfx/trie/trie.c:78: CHECK:PARENTHESIS_ALIGNMENT: Alignment should match open parenthesis
rtrlib/pfx/trie/trie.c:102: CHECK:PARENTHESIS_ALIGNMENT: Alignment should match open parenthesis
rtrlib/pfx/trie/trie.c:163: CHECK:PARENTHESIS_ALIGNMENT: Alignment should match open parenthesis
rtrlib/pfx/trie/trie.c:228: CHECK:PARENTHESIS_ALIGNMENT: Alignment should match open parenthesis
```

4.) From my side the git commits are ok now. Some general notes for the future:
- The short description of the commit message should contain the name of the module where the change took place, e.g. "pfx: change x,y", "tests: blabla", "docs: bla"
- The first line of a commit message shouldn't be longer than 50 characters, in some comits from you I have to scroll to the right to see the whole message, the first line should only be a very brief summary, everything else should be in the long description part of the commit message
- If you have in your branch a commit that fixes a bug that you introduced earlier in the same branch, I would'nt make an extra commit, I would fit it in the commit that introduced the bug before. You can e.g. use the git commit --fixup=<commit> for that and then merge those both commits with "git rebase -i".
E.g. commit " Remove trie_lookup(..) and rename trie_lookup_first_match(..) to trie_lookup(..)" could be merged with "Add trie data structure"
- git commit messages should always have a short and a long description see e.g. http://chris.beams.io/posts/git-commit/. Some examples that would have been note worthy:
  - "Add test_trie" commit -  that the file is based on test_lpfst and what changes were introduces, if there are any changes, it's not possible to get from the commit currently
  - "Fix lock placement to prevent invalid memory access in trie-pfx.c" commit - how the invalid memory access can be triggered

waehlisch:
OK, fine with me to remove LPFST.

waehlisch:
Closed because of #112.

rtrlib/pfx/trie/trie-pfx.c
fho: How can it happen that ipvX points to an invalid memory region when accessed?

If pfx_table_remove() removes the last node from the tree, it holds the lock until the ipvX pointer was set to NULL.
Afterwards the lock is obtained in pfx_table_src_remove() and it's checked if the ipvX pointer is NULL, if yes nothing is freed.

Do I miss something?

rtrlib/pfx/trie/trie-pfx.c
fho: This seems a little bit difficult.
@reuteran fix was:
```
@@ -141,8 +144,10 @@ int pfx_table_del_elem(struct node_data *data, const unsigned int index)
     }
     data->len--;
     data->ary = realloc(data->ary, sizeof(struct data_elem) * data->len);
-    if(data->len != 0 && data->ary == NULL)
+    if(data->len != 0 && data->ary == NULL) {
+        free(&data->ary[data->len]);
         return PFX_ERROR;
+    }
     else if(data->len == 0)
         data->ary = NULL;
```

The free() manpage says:

> The  free()  function frees the memory space pointed to by ptr, which must have been returned by a previous call
>        to malloc(), calloc(), or realloc().
> 

free() would be called not for the starting address of the allocated memory area. If I understand the manpage right it's a requirement that the exact passed address was previously returned by a memory allocation function.
So I think free() would fail here, maybe cause a segfault.

If realloc() fails here we would have the state that the array is 1 element bigger than we need and the unused element would be at the end.
The data->len would contain the new length already.
So to handle this situation correctly we could  restore the previous state.
This means shift the elements in the array again and insert the old value again.
I don't know if this makes so much sense.
We can also keep the current state and, the array would be 1 element bigger than needed in the very unlikely case that realloc() fails when we want to decrease the allocated memory size (not sure if this even can happen).
When the next element would be appened or deleted from the array another remalloc() try would be made. If that call would be succesful it would also fix the previous failure.

If data->len becomes 0 and we have X unused elements in the array I would expect that remalloc() can't fail because it would be the same than a call to free().

What do you think @nbars  @reuteran  @smlng?

smlng: at a first glance I agree with @fho, that `free(&data->ary[data->len]);` is wrong.

Further I think the code should check if `len == 0` **before** doing the realloc, because according to man page:

> If size is zero and ptr is not NULL, a new,
     minimum sized object is allocated and the original object is freed.

Hence we need an explicit free here if empty

smlng: or set `data->ary = NULL;` before realloc with size = 0

fho: My man page (release 4.09 of the Linux man-pages project) says something different regarding realloc() and 0 size:

>  if size is equal to zero, and ptr is not NULL, then the call
>        is equivalent to free(ptr).

Is your quote really regarding realloc() and not malloc()?
Which man page version do you have? :-)

fho: crazy, great find, I couldn't imagine that realloc() behaves so differently on Linux and BSD. :-)
I'll replace it with free() like you suggested.

fho: I think it's more consistent if the dataset doesn't change on an error.
Therefore I simply appened the element that was lost before during shifting to the end of the array again, if realloc failed

PR #104, opened 2016-12-02
lpfst: fix validation

smlng:
This is an alternative to #103 for issue #99, but it fixes the currently used prefix table struct and algorithm.

It also extends `test_pfx` and `test_live_validation` with tests to verify validation is correct.


smlng:
This doesn't fix all prevalent issues with lpfst, and fixing the remaining might not be that simple or even possible with the data structure and current lookup algorithm.

smlng:
@ColinBS updated my fix, please include into you tests again and recheck.

smlng:
wont fix

PR #105, opened 2016-12-08
protocol: do not put the received pdu in the error pdu for internal errors

mroethke:
This is in compliance with RFC6810 Section 5.10 which states that
"If the error is generic (e.g., "Internal Error") and not associated
with the PDU to which it is responding, the Erroneous PDU field MUST
be empty and the Length of Encapsulated PDU field MUST be zero."


smlng:
Looks good to me. Travis is happy, too; so ACK and merge

PR #108, opened 2017-01-08
cmake: search LIBSSH_INCLUDE, LIBSSH_LIBRARY paths for libssh files

fho:
The README file mentions that the path to the libssh library can
be specified by defining the LIBSSH_INCLUDE and LIBSSH_LIBRARY variables
but CMAKE wasn't searching in those paths for libssh.

Add those variables to the directory search list for libssh include and
lib files.

### DIFF
--- a/cmake/modules/FindLibSSH.cmake
+++ b/cmake/modules/FindLibSSH.cmake
@@ -28,6 +28,7 @@ if (NOT NoSSH)
         NAMES
           libssh/libssh.h
         PATHS
+          ${LIBSSH_INCLUDE}
           /usr/include
           /usr/local/include
           /opt/local/include
@@ -41,6 +42,7 @@ if (NOT NoSSH)
           ssh
           libssh
         PATHS
+          ${LIBSSH_LIBRARY}
           /usr/lib
           /usr/local/lib
           /opt/local/lib


cmake/modules/FindLibSSH.cmake
```diff
@@ -34,6 +34,7 @@ if (NOT NoSSH)
           /sw/include
           ${CMAKE_INCLUDE_PATH}
           ${CMAKE_INSTALL_PREFIX}/include
+          {LIBSSH_INCLUDE}
```
smlng: should be `${LIBSSH_INCLUDE}`, or not?

Further I would put those at the top (first place), because they might be used to overwrite other locations if multiple version of libssh are in the system. Same below for the library path...

cmake/modules/FindLibSSH.cmake
```diff
@@ -47,6 +48,7 @@ if (NOT NoSSH)
           /sw/lib
           ${CMAKE_LIBRARY_PATH}
           ${CMAKE_INSTALL_PREFIX}/lib
+          ${LIBSSH_LIBRARY}
```
smlng: move to top, see above comment

fho:
Thanks! I fixed both things according to your suggestions and pushed the commit again.

cmake/modules/FindLibSSH.cmake
fho: fixed

waehlisch:
All review comments fixed. Merge.

PR #112, opened 2017-02-06
Replace LPFST with SPFST to fix incorrect validation results (#99)

### LINKED ISSUE #99: Incorrect validation results

fho:
This fixes #99.

Please review, give me feedback for improvement or merge.

I verified at the weekend the results (https://github.com/rtrlib/rtrlib/issues/99#issuecomment-277433580).
They are the same than the ones of @reuteran's implementation which seems to be correct (https://github.com/rtrlib/rtrlib/issues/99#issuecomment-277687789)


smlng:
so what's the conclusion, we go for #103 or this one? As far as I understood both fix #99 -- and are essentially the same, but we only need one of them. Make a vote?

fho:
@smlng Yes
in #103 the broken LPFST is still in the code and can be reactivated with a define (but is not tested from travis for every build). 
#112 only has the new datastructure, the LPFST is removed, no define.

#103 has still some outstanding coding style issues.
#112 also has some commit that refactor the test_pfx to reduce code duplication

See the commits for the differences
https://github.com/rtrlib/rtrlib/pull/103/commits
https://github.com/rtrlib/rtrlib/pull/112/commits

smlng:
@fho I agree, that your solution is straight forward, essentially replacing a `<` with `>` when filling the table and some renaming. And replacing lpfst is the right thing to do: makes no sense to keep buggy code. 

@reuteran and @waehlisch are you okay with closing #103 in favor of this one?

waehlisch:
@smlng I don't have time to dig into the code currently. However, I suspect that the simple "s/</>/" is not sufficient. In any case, we should not mix the arguments (new Trie data structure vs. keeping or dropping LPFST, coding style ...).

As I wrote before, the final testing needs to be finished first, before we close or merge issues/PRs.
I suggest that the comparison of validation results by @ColinBS includes the SPFST.

fho:
@waehlisch
The main argument is that we only proceed very slowly with getting #104 in a mergeable state.

The changes to the datastructure in #112 and #103 are the same.
#112 and #103 change the datastructure. Nodes with shorter prefix are ordered before nodes with longer prefixes in the tree. Therefore shorter prefixes are not skipped anymore when the tree is searched. It's not about only replacing "longest" with "shortest".

Besides the theory we have also a lot of evidence from tests already:
-  all the existing validation tests in the repository are successful.
- My tests on the weekend shown that my implementation results the same than the on from @reuteran (https://github.com/rtrlib/rtrlib/issues/99#issuecomment-277433580). Also proofed again from @ColinBS (https://github.com/rtrlib/rtrlib/issues/99#issuecomment-278061812).
- The comparsion from @ColinBS indicates that the results are the same than the one from the rpki-validator now (https://github.com/rtrlib/rtrlib/issues/99#issuecomment-277687789).

So it's unlikely that the new implementation is wrong.
I agree it's better to wait for the final results, if we will have them soon. :-)
@ColinBS  is working on it?

waehlisch:
@fho, all, fine to close #103 and to merge this PR. One minor comment: SPFST is a very uncommon naming. "Trie" is much more common. Personally, I'm more in favor of call the data structure "trie" but finally I leave up to @fho.

Commit message 5e38bba3438408c9894afd186c33d20fda4fafc7 : s/They getbits/The getbits/

waehlisch:
Ah, and please consider the comments from @nbars before we merge, see #103.

fho:
@waehlisch Thanks, I fixed the getbits typo.
Ups, Yes I overlooked the reports from @nbars, thought they were introduced in #103 but the bugs are indeed already present in the master version.
I created #104 to fix them. I will do it latest on the weekend (if nobody else did it until then :-)).

For me both names would be fine. Trie seems more general, spfst says a little bit more about the tree structure but nobody knows what the spfst abbreviation means. :-)
@smlng  @reuteran  do you have an opinion regarding the naming?

smlng:
@fho, I agree that _spfst_ (if one knows how to resolve the acronym) is more speaking (?) than _trie_. Further, we used _lpfst_ so far and nobody wondered or complaint about it, so why not _spfst_: 👍 from my side?

fho:
@smlng for LPFST there is at least one paper that describes the algorithm and uses the same abbreviation (https://pdfs.semanticscholar.org/ef30/f4ae3e33b69688661d009385d24295e1ce1f.pdf) :-)
Ok, than lets keep spfst it and if we like we can still change it later.

@smlng if you don't have any further suggestions for improvements, can you merge it to master or should I do it?

waehlisch:
Just to be clear, I'm much more in favor of changing the name to. No book etc. knows the term SPFST.

fho:
I changed the name to trie.

waehlisch:
Thanks, in particular @fho and @reuteran!

smlng:
why is this closed without merge? Did I miss something?

waehlisch:
Strange. I thought it was merged.

Issue #114, opened 2017-02-08
Fix pfx-lpst/spfst issues found by @nbars


fho:
@nbars  pointed out multiple issues in the pfx-lpst/spfst implementation that already exist in the master branch.
@reuteran fixed the issues already in his branch, cherry-pick his fixes if possible.

fho:
I wrote 2 comments in the #103 issue regarding the initial fixes:
https://github.com/rtrlib/rtrlib/pull/103#discussion_r100516074
https://github.com/rtrlib/rtrlib/pull/103#discussion_r100518018

PR #117, opened 2017-02-10
Fix possible memory leak in pfx_table_append_elem()

fho:
fixes issues mentioned in https://github.com/rtrlib/rtrlib/issues/114

### DIFF
--- a/rtrlib/pfx/trie/trie-pfx.c
+++ b/rtrlib/pfx/trie/trie-pfx.c
@@ -153,18 +153,31 @@ inline struct trie_node *pfx_table_get_root(const struct pfx_table *pfx_table, c
 
 int pfx_table_del_elem(struct node_data *data, const unsigned int index)
 {
+    struct data_elem *tmp;
+    struct data_elem deleted_elem = data->ary[index];
+
     //if index is not the last elem in the list, move all other elems backwards in the array
     if(index != data->len - 1) {
         for(unsigned int i = index; i < data->len - 1; i++) {
             data->ary[i] = data->ary[i+1];
         }
     }
+
     data->len--;
-    data->ary = realloc(data->ary, sizeof(struct data_elem) * data->len);
-    if(data->len != 0 && data->ary == NULL)
-        return PFX_ERROR;
-    else if(data->len == 0)
+    if (!data->len) {
+        free(data->ary);
         data->ary = NULL;
+        return PFX_SUCCESS;
+    }
+
+    tmp = realloc(data->ary, sizeof(struct data_elem) * data->len);
+    if (!tmp) {
+        data->ary[data->len] = deleted_elem;
+        data->len++;
+        return PFX_ERROR;
+    }
+
+    data->ary = tmp;
 
     return PFX_SUCCESS;
 }
--- a/rtrlib/pfx/trie/trie-pfx.c
+++ b/rtrlib/pfx/trie/trie-pfx.c
@@ -186,7 +186,8 @@ int pfx_table_add(struct pfx_table *pfx_table, const struct pfx_record *record)
             //append record to note_data array
             int rtval = pfx_table_append_elem(node->data, record);
             pthread_rwlock_unlock(&pfx_table->lock);
-            pfx_table_notify_clients(pfx_table, record, true);
+            if (rtval == PFX_SUCCESS)
+                pfx_table_notify_clients(pfx_table, record, true);
             return rtval;
         } else {
             //no node with same prefix and prefix_len found
--- a/rtrlib/pfx/trie/trie-pfx.c
+++ b/rtrlib/pfx/trie/trie-pfx.c
@@ -83,21 +83,26 @@ void pfx_table_free(struct pfx_table *pfx_table)
 
 int pfx_table_append_elem(struct node_data *data, const struct pfx_record *record)
 {
-    data->len++;
-    data->ary = realloc(data->ary, sizeof(struct data_elem) * data->len);
-    if(data->ary == NULL)
+    struct data_elem *tmp  = realloc(data->ary, sizeof(struct data_elem) * ((data->len) + 1));
+    if(tmp == NULL)
         return PFX_ERROR;
+    data->len++;
+    data->ary = tmp;
     data->ary[data->len - 1].asn = record->asn;
     data->ary[data->len - 1].max_len = record->max_len;
     data->ary[data->len - 1].socket = record->socket;
     return PFX_SUCCESS;
 }
 
-int pfx_table_create_node(struct trie_node **node, const struct pfx_record *record)
+int pfx_table_create_node(struct trie_node **node,
+                          const struct pfx_record *record)
 {
+    int err;
+
     *node = malloc(sizeof(struct trie_node));
     if(*node == NULL)
         return PFX_ERROR;
+
     (*node)->prefix = record->prefix;
     (*node)->len = record->min_len;
     (*node)->lchild = NULL;
@@ -105,12 +110,27 @@ int pfx_table_create_node(struct trie_node **node, const struct pfx_record *reco
     (*node)->parent = NULL;
 
     (*node)->data = malloc(sizeof(struct node_data));
-    if((*node)->data == NULL)
-        return PFX_ERROR;
+    if((*node)->data == NULL) {
+        err = PFX_ERROR;
+        goto free_node;
+    }
+
     ((struct node_data *) (*node)->data)->len = 0;
     ((struct node_data *) (*node)->data)->ary = NULL;
 
-    return pfx_table_append_elem(((struct node_data *) (*node)->data), record);
+    err = pfx_table_append_elem(((struct node_data *) (*node)->data), record);
+    if (err)
+        goto free_node_data;
+
+    return PFX_SUCCESS;
+
+
+free_node_data:
+    free((*node)->data);
+free_node:
+    free(*node);
+
+    return err;
 }
 
 struct data_elem *pfx_table_find_elem(const struct node_data *data, const struct pfx_record *record, unsigned int *index)


rtrlib/pfx/trie/trie-pfx.c
```diff
@@ -83,34 +83,54 @@ void pfx_table_free(struct pfx_table *pfx_table)
 
 int pfx_table_append_elem(struct node_data *data, const struct pfx_record *record)
 {
-    data->len++;
-    data->ary = realloc(data->ary, sizeof(struct data_elem) * data->len);
-    if(data->ary == NULL)
+    struct data_elem *tmp  = realloc(data->ary, sizeof(struct data_elem) * ((data->len) + 1));
+    if(tmp == NULL)
```
smlng: coding style: insert space before parentheses

rtrlib/pfx/trie/trie-pfx.c
```diff
@@ -83,34 +83,54 @@ void pfx_table_free(struct pfx_table *pfx_table)
...
 }
 
-int pfx_table_create_node(struct trie_node **node, const struct pfx_record *record)
+int pfx_table_create_node(struct trie_node **node,
+			  const struct pfx_record *record)
 {
+    int err;
+
     *node = malloc(sizeof(struct trie_node));
     if(*node == NULL)
```
smlng: coding style, again

rtrlib/pfx/trie/trie-pfx.c
```diff
@@ -83,34 +83,54 @@ void pfx_table_free(struct pfx_table *pfx_table)
...
     (*node)->len = record->min_len;
     (*node)->lchild = NULL;
     (*node)->rchild = NULL;
     (*node)->parent = NULL;
 
     (*node)->data = malloc(sizeof(struct node_data));
-    if((*node)->data == NULL)
-        return PFX_ERROR;
+    if((*node)->data == NULL) {
+			err = PFX_ERROR;
```
smlng: coding style: check indention

rtrlib/pfx/trie/trie-pfx.c
```diff
@@ -83,34 +83,54 @@ void pfx_table_free(struct pfx_table *pfx_table)
...
     (*node)->prefix = record->prefix;
     (*node)->len = record->min_len;
     (*node)->lchild = NULL;
     (*node)->rchild = NULL;
     (*node)->parent = NULL;
 
     (*node)->data = malloc(sizeof(struct node_data));
-    if((*node)->data == NULL)
-        return PFX_ERROR;
+    if((*node)->data == NULL) {
```
smlng: could be shortened to 
```
if (((*node)->data = malloc(..)) == NULL) {
```

rtrlib/pfx/trie/trie-pfx.c
```diff
@@ -83,34 +83,54 @@ void pfx_table_free(struct pfx_table *pfx_table)
...
 
-    return pfx_table_append_elem(((struct node_data *) (*node)->data), record);
+    err = pfx_table_append_elem(((struct node_data *) (*node)->data), record);
+    if (err)
+	    goto free_node_data;
+
+    return PFX_SUCCESS;
+
+
+free_node_data:
```
smlng: I still don't like `goto` and IMHO they are not needed here. However, I can propose to refactor in a separate PR

rtrlib/pfx/trie/trie-pfx.c
smlng: and: coding style

rtrlib/pfx/trie/trie-pfx.c
```diff
@@ -83,34 +83,54 @@ void pfx_table_free(struct pfx_table *pfx_table)
...
+    free((*node)->data);
+free_node:
+	free(*node);
```
smlng: check indention here and above at `free()`

smlng:
please add `trie_pfx.c` to the coding style check

rtrlib/pfx/trie/trie-pfx.c
fho: I still like it a lot. It makes error handling much more readable and easier to follow then thousands of nested ifs :-)

fho:
@smlng trie-pfx.c was never adapted to the kernel coding style. I don't want to adapt the whole file in this branch and add it to check-coding-files.txt. That would a be seperate task.
I fixed the spacing inconsistency.

rtrlib/pfx/trie/trie-pfx.c
smlng: I agree, iff there are lots of error cases (N > 2, 😉) that have to do similar cleanup operations before return. However, here we have just 2 error cases; I admit, one is reusing (or requires) the same free() as the other. 

IMHO this could be written nicely with 2 nested`if () else`. However, you may leave as is - just my personal view.

fho: @smlng it's even adviced in the kernel coding style guide:
https://github.com/torvalds/linux/blob/master/Documentation/process/coding-style.rst see "7) Centralized exiting of functions" :-)

smlng:
@fho: I thought we had the (in)official rule: if one touches code or a file, it has to comply with Kernel coding style afterwards 😎

rtrlib/pfx/trie/trie-pfx.c
fho: I prefer it in 2 lines. I think it's better to read.
whatever, minor personal preferences I think :-)

fho:
@smlng oh god, I hope we didn't  :-)

rtrlib/pfx/trie/trie-pfx.c
smlng: Fine with me

fho:
I added another commit to fix the realloc() problems in pfx_table_del_elem() that we discussed in #103

rtrlib/pfx/trie/trie-pfx.c
fho: Yes, doesn't make a huge difference in this case with only 2 error handlers :-)

rtrlib/pfx/trie/trie-pfx.c
```diff
@@ -133,18 +153,31 @@ inline struct trie_node *pfx_table_get_root(const struct pfx_table *pfx_table, c
 
 int pfx_table_del_elem(struct node_data *data, const unsigned int index)
 {
+    struct data_elem *tmp;
+    struct data_elem deleted_elem = data->ary[index];
```
smlng: what if `index > data->len -1`?

smlng: I mean, if this is possible, anyway?

rtrlib/pfx/trie/trie-pfx.c
```diff
@@ -133,18 +153,31 @@ inline struct trie_node *pfx_table_get_root(const struct pfx_table *pfx_table, c
...
-    else if(data->len == 0)
+    if (!data->len) {
+        free(data->ary);
         data->ary = NULL;
+        return PFX_SUCCESS;
+    }
+
+    tmp = realloc(data->ary, sizeof(struct data_elem) * data->len);
+    if (!tmp) {
+        data->ary[data->len] = deleted_elem;
```
smlng: I'm not sure if this is necessary, as I understood the man page the memory remains unchanged on error, hence the element should still be in place and reset not required?

fho: It is, except when the element that should be deleted is the last element in the list. In that case it would be redundant. I think the possible redundant operation doesn't really matter, that realloc() fails is very unlikely already. So I didn't add another if-statement for the case, to keep it more simple instead.

If the element that should be deleted isn't at the end of the list, all elements in the list gets shifted to the left one position.
The data->ary[index] is oververwritten with the element at data->ary[index + 1].
The element at data->ary[data->len - 2] is therefore the same than the one at data->ary[data->len - 1]. 
So I saved the data->ary[index] elem and insert it at data->ary[data->len - 1] if realloc() failed.

smlng: You are right, oversaw this (very common) case -> my bad ...

rtrlib/pfx/trie/trie-pfx.c
fho: It shouldn't be. The function requires that you give a valid index value :-)

rtrlib/pfx/trie/trie-pfx.c
fho: @smlng btw when reading the realloc() manpages I discovered reallocf(), seems to also exist on Linux and it frees the memory when realloc() fails, sounds very handy

rtrlib/pfx/trie/trie-pfx.c
smlng: where is this ensured? And if thats the case, we should have an `assert((data->len == 0) || (index < date->len))` here at least, or not?

rtrlib/pfx/trie/trie-pfx.c
smlng: sounds good, if that is available on any Linux and UNIX platform, I'm in ...

rtrlib/pfx/trie/trie-pfx.c
smlng: well I know that C doesn't protect you from doing something stupid.

Nevertheless, the assert might come in handy for debugging. IMO an assert also makes it clear which values are expected for parameters. Hence, `assert(data != NULL)` could be added here, too.

However, we might discuss this separately - IMO asserts are useful and we should think about adding more often throughout the code. In a release build they are stripped anyway, so no code bloat ...

fho: Yes, I totally agree that asserts() can be very useful when debugging to find the source of an error.

fho:
@smlng can the branch be merged?

smlng:
Didn't you want to rewrite to use `reallocf`, and address some coding style issue, i.e. space before parentheses. I don't see those changes in the commits?!

fho:
No, I think it's better to keep one  coding style in the file than introducing a second style for a few functions in that file.

reallocf() was only a find that I made which I  shared. I don't see that it would benefit the bugfixing here.

The issue/branch is about fixing the bugs that were found not about refactoring the file.

smlng:
okay, just the fixes then

smlng:
oh and please squash

fho:
Why you would like to squash 3 commits that are fixing different issues into one commit?

It would make it hard to understand what code change fixes which specific issue and would also make it impossible to revert one of the fixes seperately if they turned out to be wrong.

smlng:
mhm, I thought of this PR as *one* task fixing some issues in `trie-pfx.c`, which would be one commit (for me). However, if you feel better with 3 separate commits, merge as is - its fine, too.

Issue #118, opened 2017-03-02
Add integration readthedocs


waehlisch:
Allow for our Doxygen doc into ReadTheDocs, see for example: https://github.com/Cruel/readthedocs-breathe/tree/master/docs

PR #119, opened 2017-03-28
docs: add readthedocs doxygen wrapper

smlng:
addressed #118 

This may need some work. Currently it shows (lots of) errors during generation, and the output is single-page -> hence, a bit unstructured.


waehlisch:
@smlng thanks. Can you provide a readthedocs link to a preview? When do you will fix the errors?

Btw, now we have two Doxyfile, which might be confusing. Should be cleaned up later.

smlng:
Regarding the 2 Doxyfiles: I'm not sure if they can be merged into one, the problem is that `doxygen/Doxyfile` is used by the Cmake toolchain and `docs/Doxyfile` is for the RTD intgeration, which needs some adaptions for output params and paths.

smlng:
@waehlisch I put a demo side online, have a look [here](http://rtrlib-rtd.inet.haw-hamburg.de).

Regarding the errors, I see during doc generation: most of them are related to undocumented code parts (functions, parameters, and so on), some are readthedocs + breathe related, but not severe.

PR #122, opened 2017-04-20
packets: fix potential double free vulnerabilities

mroethke:
In case of a realloc failure during processing of new prefixes and router
keys the arrays where the new data is stored after initial processing are
freed twice. The first time in the processing functions rtr_store_router_key_pdu
or rtr_store_prefix_pdu, depending on the data, and during cleanup in
rtr_sync_receive_and_store_pdus which calls the earlier mentioned ones.

coverity ids: 1420788, 1420796, 1420797

### DIFF
--- a/rtrlib/rtr/packets.c
+++ b/rtrlib/rtr/packets.c
@@ -749,6 +749,10 @@ static int rtr_undo_update_spki_table(struct rtr_socket *rtr_socket, void *pdu)
 
 /*
  * @brief Appends the Prefix PDU pdu to ary.
+ *
+ * @return RTR_SUCCESS On success
+ * @return RTR_ERROR On realloc failure
+ * @attention ary is not freed in this case, because it might contain data that is still needed
  */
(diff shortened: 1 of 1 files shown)


rtrlib/rtr/packets.c
```diff
@@ -749,6 +749,9 @@ static int rtr_undo_update_spki_table(struct rtr_socket *rtr_socket, void *pdu)
 
 /*
  * @brief Appends the Prefix PDU pdu to ary.
+ *
+ * @retval RTR_SUCCESS On success
+ * @retval RTR_ERROR On realloc failure, ary is not freed in this case, because it might contain data that is still needed
```
smlng: for sake of consistency use `@return` here, like in the rest of the docu.

Though, we might reconsider the use of `@retval` and `@return` for docu, in general, but that would be another PR.

mroethke:
I changed the documentation accordingly.

PR #126, opened 2017-05-22
rtr_mgr: fix resource leaks in rtr_mgr_init

smlng:
follow up on #124 

coverty ids: 1420783, 1420791

### DIFF
--- a/rtrlib/rtr_mgr.c
+++ b/rtrlib/rtr_mgr.c
@@ -293,6 +293,8 @@ int rtr_mgr_init(struct rtr_mgr_config **config_out,
 	struct pfx_table *pfxt = NULL;
 	struct spki_table *spki_table = NULL;
 	struct rtr_mgr_config *config = NULL;
+	uint8_t last_preference = UINT8_MAX;
+
 	*config_out = NULL;
 
 	if (groups_len == 0) {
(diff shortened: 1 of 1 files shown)


smlng:
@mroethke can you please check this PR with coverity, to see if the defects are fixed? Thx ...

rtrlib/rtr_mgr.c
```diff
@@ -293,7 +293,7 @@ int rtr_mgr_init(struct rtr_mgr_config **config_out,
 	struct pfx_table *pfxt = NULL;
 	struct spki_table *spki_table = NULL;
 	struct rtr_mgr_config *config = NULL;
-	*config_out = NULL;
```
mroethke: I think we should keep setting *config_out = NULL, because if groups_len = 0 the value of *config_out is now undefined, but should be NULL according the doxygen comment.

smlng: mhm, okay I'll re-add that - to be double sure 😉 

IMHO it should be the responsibility of the calling function to ensure correct usage and initialization of `config_out`.

smlng:
fixed and directly squashed. @mroethke : please re-run coverity, to verify that this fixes named issues and also does not introduce new ones.

PR #129, opened 2017-06-06
lib: fix ip docu

smlng:
fixes doxygen documentation in `lib/ipv4.h` and `lib/ipv6.h` and also adapt coding style for the latter.


smlng:
I'd like to have at least one review/opinion before merge: @mroethke?!

PR #130, opened 2017-07-26
Alloc utils

reuteran:
This is the malloc wrapper written by @mroethke at the IETF99 Hackathon. This was originally in the dynamic-groups branch, but I've separated it out.

### DIFF
--- a/rtrlib/lib/alloc_utils.c
+++ b/rtrlib/lib/alloc_utils.c
@@ -11,9 +11,9 @@
 
 #include "alloc_utils.h"
 
-void* (*MALLOC_PTR)(size_t size);
-void* (*REALLOC_PTR)(void* ptr, size_t size);
-void (*FREE_PTR)(void* ptr);
+void* (*MALLOC_PTR)(size_t size) = malloc;
+void* (*REALLOC_PTR)(void* ptr, size_t size) = realloc;
+void (*FREE_PTR)(void* ptr) = free;
 
 
 void lrtr_set_malloc(void* (*malloc_function)(size_t size))
@@ -33,27 +33,15 @@ void lrtr_set_free(void (free_ptr)(void* ptr))
 
 inline void* lrtr_malloc(size_t size)
 {
-	if (MALLOC_PTR) {
-		return MALLOC_PTR(size);
-	} else {
-		return malloc(size);
-	}
+	return MALLOC_PTR(size);
 }
 
 inline void lrtr_free(void* ptr)
 {
-	if (FREE_PTR) {
-		return FREE_PTR(ptr);
-	} else {
-		return free(ptr);
-	}
+	return FREE_PTR(ptr);
 }
 
 inline void* lrtr_realloc(void* ptr, size_t size)
 {
-	if (REALLOC_PTR) {
-		return REALLOC_PTR(ptr, size);
-	}  else {
-		return realloc(ptr, size);
-	}
+	return REALLOC_PTR(ptr, size);
 }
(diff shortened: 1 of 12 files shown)


nbars:
Could you state why we need something like this?

mroethke:
It allows better integration in other projects like frr. 
They use a custom macro that enables them to track memory usage. 
Memory allocated by rtrlib could not be tracked without this wrapper

rtrlib/lib/alloc_utils.c
```diff
@@ -0,0 +1,59 @@
...
+ * See the file LICENSE in the top level directory for more details.
+ *
+ * Website: http://rtrlib.realmv6.org/
+ */
+
+#include <stdlib.h>
+
+#include "alloc_utils.h"
+
+void* (*MALLOC_PTR)(size_t size);
```
nbars: How about changing this to ```void* (*MALLOC_PTR)(size_t size) = malloc;```?
Then we could skip the checks in ```lrtr_malloc``` and ```lrtr_free```.

waehlisch:
Seems that we only one open suggestion by @nbars, right? Anything else?

rtrlib/lib/alloc_utils.c
mroethke: you only mention malloc and free, is there a reason why this would not work for realloc?

nbars: No, just forgot to mention ```realloc```.

mroethke:
@nbars thank you for the suggestion. I changed it accordingly.

rtrlib/lib/alloc_utils.h
```diff
@@ -0,0 +1,46 @@
...
+extern void *(*MALLOC_PTR)(size_t size);
+extern void (*FREE_PTR)(void *ptr);
+extern void *(*REALLOC_PTR)(void *ptr, size_t size);
+
+/**
+ * @brief Sets a custom malloc function that is used throughout rtrlib.
+ * If you set this function you must also set custom free and custom realloc
+ * @param[in] Pointer to a malloc function
+ */
+void lrtr_set_malloc(void *(*malloc_function)(size_t size));
```
fho: If always all functions has to be set, an alternative could be to have a single register() function that takes 3 function pointers or a struct that contains the function pointers.
I don't know other libraries implementing it, maybe there is a common pattern for it that we can follow (if we are not doing it already).

rtrlib/lib/alloc_utils.c
```diff
@@ -0,0 +1,46 @@
...
+ * Website: http://rtrlib.realmv6.org/
+ */
+
+#include <stdlib.h>
+
+#include "alloc_utils.h"
+
+void *(*MALLOC_PTR)(size_t size) = malloc;
+void *(*REALLOC_PTR)(void *ptr, size_t size) = realloc;
+void (*FREE_PTR)(void *ptr) = free;
```
fho: The uppercase hurts my eyes :-D

rtrlib/lib/alloc_utils.h
mroethke: It did not find much, probably due to being hard to search for not because few people do this, but those that I did find indeed use just one function for this.

I will change it accordingly.

rtrlib/lib/alloc_utils.c
mroethke: Now that I think about it since they are not constants a lowercase name would fit and of course we do not want your eyes to hurd ;)

ColinBS:
If there are no more major issues about this PR, I would like to consider it for merging. This PR is required to finish up PR #131 . As soon as PR #131 is merged, the FRR integration can progress.

smlng:
I don't see any imminent issues either, if no-one objects until tomorrow (16.08.17) morning I'll merge this. Regarding UPPERCASE or not, this should not block general progress and improvements of the library - and can be fixed/adapted later on if needed.

PR #131, opened 2017-07-26
Dynamic adding and removing of rtr_mgr_groups

reuteran:
This uses a linked list instead of an array to store rtr_mgr_groups. It also adds two new functions for the user: rtr_mgr_group_add and rtr_mgr_group_remove. These allow users to add and remove groups while rtr_mgr is running. 

Right now this uses a tommy_list out of convenience. This is a generic doubly linked list. 
This means there are now calls to tommy_* functions in rtr_mgr.c 
Do we want to write our own linked list? Or some kind of wrapper for tommy_list?

Since I am in the middle of writing my masters thesis, it would be good if @ColinBS and @mroethke could pick this up and add necessary changes.  This should include more extensive tests, such as adding/removing groups while rtr_mgr is running and checking whether the changes work as they should there as well.

### DIFF
--- a/rtrlib/lib/alloc_utils.h
+++ b/rtrlib/lib/alloc_utils.h
@@ -36,10 +36,11 @@ void *lrtr_realloc(void *ptr, size_t size);
 
 /**
  * @brief Duplicates a string
+ * @pre string != NULL
  * @param[in] string
  * @returns Duplicated string
  * @returns NULL on error
  */
-char *lrtr_strdup (const char *string);
+char *lrtr_strdup(const char *string);
 
 #endif
--- a/rtrlib/lib/alloc_utils.h
+++ b/rtrlib/lib/alloc_utils.h
@@ -34,4 +34,12 @@ void lrtr_free(void *ptr);
 
 void *lrtr_realloc(void *ptr, size_t size);
 
+/**
+ * @brief Duplicates a string
+ * @param[in] string
+ * @returns Duplicated string
+ * @returns NULL on error
+ */
+char *lrtr_strdup (const char *string);
+
 #endif
--- a/rtrlib/rtr_mgr.c
+++ b/rtrlib/rtr_mgr.c
@@ -27,11 +27,14 @@ static const char * const mgr_str_status[] = {
 	[RTR_MGR_ERROR] = "RTR_MGR_ERROR",
 };
 
-static int rtr_mgr_find_group(const struct rtr_mgr_config *config,
-			      const struct rtr_socket *sock,
-			      unsigned int *ind);
+static struct rtr_mgr_group *rtr_mgr_find_group(struct rtr_mgr_config *config,
+						const struct rtr_socket *sock);
 static int rtr_mgr_config_cmp(const void *a, const void *b);
-static bool rtr_mgr_config_status_is_synced(const struct rtr_mgr_group *config);
+static int rtr_mgr_config_cmp_tommy(const void *a, const void *b);
+static bool rtr_mgr_config_status_is_synced(const struct rtr_mgr_group *group);
+static void rtr_mgr_cb(const struct rtr_socket *sock,
+		       const enum rtr_socket_state state,
+		       void *data_config, void *data_group);
 
 static void set_status(const struct rtr_mgr_config *conf,
 		       struct rtr_mgr_group *group,
@@ -59,21 +62,48 @@ static int rtr_mgr_start_sockets(struct rtr_mgr_group *group)
 	return RTR_SUCCESS;
 }
 
-int rtr_mgr_find_group(const struct rtr_mgr_config *config,
-		       const struct rtr_socket *sock,
-		       unsigned int *ind)
+static int rtr_mgr_init_sockets(struct rtr_mgr_group *group,
+				struct rtr_mgr_config *config,
+				const unsigned int refresh_interval,
+				const unsigned int expire_interval,
+				const unsigned int retry_interval)
 {
-	for (unsigned int i = 0; i < config->len; i++) {
-		for (unsigned int j = 0;
-		     j < config->groups[i].sockets_len; j++) {
-			if (config->groups[i].sockets[j] == sock) {
-				*ind = i;
-				return RTR_SUCCESS;
+	for (unsigned int i = 0; i < group->sockets_len; i++) {
+		enum rtr_rtvals err_code = rtr_init(group->sockets[i], NULL,
+						    config->pfx_table,
+						    config->spki_table,
+						    refresh_interval,
+						    expire_interval,
+						    retry_interval,
+						    rtr_mgr_cb, config, group);
+		if (err_code)
+			return err_code;
+	}
+	return RTR_SUCCESS;
+}
+
+static struct rtr_mgr_group *rtr_mgr_find_group(struct rtr_mgr_config *config,
+						const struct rtr_socket *sock)
+{
+	tommy_node *node;
+	struct rtr_mgr_group_node *group_node;
+
+	pthread_mutex_lock(&config->mutex);
+	node  = tommy_list_head(&config->groups);
+
+	while (node) {
+		group_node = node->data;
+		for (unsigned int j = 0; j < group_node->group->sockets_len;
+									 j++) {
+			if (group_node->group->sockets[j] == sock) {
+				pthread_mutex_unlock(&config->mutex);
+				return group_node->group;
 			}
 		}
+	node = node->next;
 	}
-	MGR_DBG1("Error couldn't find a wanted rtr_socket in rtr_mgr_config");
-	return RTR_ERROR;
+	pthread_mutex_unlock(&config->mutex);
+	return NULL;
 }
 
 bool rtr_mgr_config_status_is_synced(const struct rtr_mgr_group *group)
@@ -91,135 +121,170 @@ bool rtr_mgr_config_status_is_synced(const struct rtr_mgr_group *group)
 
 static void rtr_mgr_close_less_preferable_groups(const struct rtr_socket *sock,
 						 struct rtr_mgr_config *config,
-						 unsigned int my_group_idx)
+						 struct rtr_mgr_group *group)
 {
-	for (unsigned int i = 0; i < config->len; i++) {
-		struct rtr_mgr_group cg = config->groups[i];
+	struct rtr_mgr_group_node *group_node;
+	struct rtr_mgr_group *current_group;
 
-		if ((cg.status != RTR_MGR_CLOSED) && (i != my_group_idx) &&
-		    (cg.preference > config->groups[my_group_idx].preference)) {
-			for (unsigned int j = 0; j < cg.sockets_len; j++) {
-				pthread_mutex_unlock(&config->mutex);
-				rtr_stop(cg.sockets[j]);
-				pthread_mutex_lock(&config->mutex);
+	pthread_mutex_lock(&config->mutex);
+	tommy_node *node = tommy_list_head(&config->groups);
+
+	while (node) {
+		group_node = node->data;
+		current_group = group_node->group;
+		if ((current_group->status != RTR_MGR_CLOSED) &&
+		    (current_group != group) &&
+		    (current_group->preference > group->preference)) {
+			for (unsigned int j = 0;
+				j < current_group->sockets_len; j++) {
+				rtr_stop(current_group->sockets[j]);
 			}
-			set_status(config, &cg, RTR_MGR_CLOSED, sock);
+			set_status(config, current_group, RTR_MGR_CLOSED, sock);
 		}
+	node = node->next;
 	}
+	pthread_mutex_unlock(&config->mutex);
 }
 
-static struct rtr_mgr_group
-*get_best_inactive_rtr_mgr_group(struct rtr_mgr_config *config,
-				 unsigned int my_group_idx)
+static struct rtr_mgr_group *get_best_inactive_rtr_mgr_group(
+						struct rtr_mgr_config *config,
+						struct rtr_mgr_group *group)
 {
-	/* groups are sorted by preference */
-	for (int i = 0; i < config->len; i++) {
-		if ((i != my_group_idx) &&
-		    (config->groups[i].status == RTR_MGR_CLOSED))
-			return &config->groups[i];
+	struct rtr_mgr_group_node *group_node;
+	struct rtr_mgr_group *current_group;
+
+	pthread_mutex_lock(&config->mutex);
+	tommy_node *node = tommy_list_head(&config->groups);
+
+	while (node) {
+		group_node = node->data;
+		current_group = group_node->group;
+			if ((current_group != group) &&
+			    (current_group->status == RTR_MGR_CLOSED)) {
+				pthread_mutex_unlock(&config->mutex);
+				return current_group;
+			}
+		node = node->next;
 	}
+		pthread_mutex_unlock(&config->mutex);
 	return NULL;
 }
 
 static bool is_some_rtr_mgr_group_established(struct rtr_mgr_config *config)
 {
-	for (int i = 0; i < config->len; i++) {
-		if (config->groups[i].status == RTR_MGR_ESTABLISHED)
+	struct rtr_mgr_group_node *group_node;
+
+	pthread_mutex_lock(&config->mutex);
+	tommy_node *node = tommy_list_head(&config->groups);
+
+	while (node) {
+		group_node = node->data;
+		if (group_node->group->status == RTR_MGR_ESTABLISHED) {
+			pthread_mutex_unlock(&config->mutex);
 			return true;
+		}
+		node = node->next;
 	}
+	pthread_mutex_unlock(&config->mutex);
 	return false;
 }
 
 static inline void _rtr_mgr_cb_state_shutdown(const struct rtr_socket *sock,
 					      struct rtr_mgr_config *config,
-					      unsigned int ind)
+					      struct rtr_mgr_group *group)
 {
 	bool all_down = true;
 
-	for (unsigned int i = 0; i < config->groups[ind].sockets_len; i++) {
-		if (config->groups[ind].sockets[i]->state != RTR_SHUTDOWN) {
+	for (unsigned int i = 0; i < group->sockets_len; i++) {
+		if (group->sockets[i]->state != RTR_SHUTDOWN) {
 			all_down = false;
 			break;
 		}
 	}
 	if (all_down)
-		set_status(config, &config->groups[ind],
+		set_status(config, group,
 			   RTR_MGR_CLOSED, sock);
 	else
-		set_status(config, &config->groups[ind],
-			   config->groups[ind].status, sock);
+		set_status(config, group,
+			   group->status, sock);
 }
 
 static inline void _rtr_mgr_cb_state_established(const struct rtr_socket *sock,
 						 struct rtr_mgr_config *config,
-						 unsigned int ind)
+						 struct rtr_mgr_group *group)
 {
 	/* socket established a connection to the rtr_server */
-	if (config->groups[ind].status == RTR_MGR_CONNECTING) {
+	if (group->status == RTR_MGR_CONNECTING) {
 		/*
 		 * if previous state was CONNECTING, check if all
 		 * other sockets in the group also have a established
 		 * connection, if yes change group state to ESTABLISHED
 		 */
-		if (rtr_mgr_config_status_is_synced(&config->groups[ind])) {
-			set_status(config, &config->groups[ind],
+		if (rtr_mgr_config_status_is_synced(group)) {
+			set_status(config, group,
 				   RTR_MGR_ESTABLISHED, sock);
-			rtr_mgr_close_less_preferable_groups(sock, config, ind);
+			rtr_mgr_close_less_preferable_groups(sock, config,
+							     group);
 		} else {
-			set_status(config, &config->groups[ind],
+			set_status(config, group,
 				   RTR_MGR_CONNECTING, sock);
 		}
-	} else if (config->groups[ind].status == RTR_MGR_ERROR) {
-		/*
-		 * if previous state was ERROR, only change state to
+	} else if (group->status == RTR_MGR_ERROR) {
+		/* if previous state was ERROR, only change state to
 		 * ESTABLISHED if all other more preferable socket
 		 * groups are also in ERROR or SHUTDOWN state
 		 */
 		bool all_error = true;
+		struct rtr_mgr_group_node *group_node;
+		struct rtr_mgr_group *current_group;
 
-		for (unsigned int i = 0; (i < config->len) && all_error; i++) {
-			struct rtr_mgr_group cg = config->groups[i];
+		pthread_mutex_lock(&config->mutex);
+		tommy_node *node = tommy_list_head(&config->groups);
 
-			if (i != ind &&
-			    cg.status != RTR_MGR_ERROR &&
-			    cg.status != RTR_MGR_CLOSED &&
-			    cg.preference < config->groups[ind].preference)
+		while (node) {
+			group_node = node->data;
+			current_group = group_node->group;
+			if ((current_group != group) &&
+			    (current_group->status != RTR_MGR_ERROR) &&
+			    (current_group->status != RTR_MGR_CLOSED) &&
+			    (current_group->preference < group->preference)) {
 				all_error = false;
+			}
+			node = node->next;
 		}
-		if (all_error &&
-		    rtr_mgr_config_status_is_synced(&config->groups[ind])) {
-			set_status(config, &config->groups[ind],
-				   RTR_MGR_ESTABLISHED, sock);
-			rtr_mgr_close_less_preferable_groups(sock, config, ind);
+		pthread_mutex_unlock(&config->mutex);
+
+		if (all_error && rtr_mgr_config_status_is_synced(group)) {
+			set_status(config, group, RTR_MGR_ESTABLISHED, sock);
+			rtr_mgr_close_less_preferable_groups(sock, config,
+							     group);
 		} else {
-			set_status(config, &config->groups[ind],
-				   RTR_MGR_ERROR, sock);
+			set_status(config, group, RTR_MGR_ERROR, sock);
 		}
 	}
 }
 
 static inline void _rtr_mgr_cb_state_connecting(const struct rtr_socket *sock,
 						struct rtr_mgr_config *config,
-						unsigned int ind)
+						struct rtr_mgr_group *group)
 {
-	if (config->groups[ind].status == RTR_MGR_ERROR)
-		set_status(config, &config->groups[ind],
+	if (group->status == RTR_MGR_ERROR)
+		set_status(config, group,
 			   RTR_MGR_ERROR, sock);
 	else
-		set_status(config, &config->groups[ind],
+		set_status(config, group,
 			   RTR_MGR_CONNECTING, sock);
 }
 
 static inline void _rtr_mgr_cb_state_error(const struct rtr_socket *sock,
 					   struct rtr_mgr_config *config,
-					   unsigned int ind)
+					   struct rtr_mgr_group *group)
 {
-	set_status(config, &config->groups[ind],
-		   RTR_MGR_ERROR, sock);
+	set_status(config, group, RTR_MGR_ERROR, sock);
 
 	if (!is_some_rtr_mgr_group_established(config)) {
 		struct rtr_mgr_group *next_group =
-			get_best_inactive_rtr_mgr_group(config, ind);
+			get_best_inactive_rtr_mgr_group(config, group);
 
 		if (next_group)
 			rtr_mgr_start_sockets(next_group);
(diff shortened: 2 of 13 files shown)


rtrlib/rtr_mgr.h
```diff
@@ -67,28 +68,37 @@ struct rtr_mgr_group {
 	enum rtr_mgr_status status;
 };
 
+//TODO Find a nicer way todo a linked list (without writing our own)
```
fho: I recommend to use the linked list from the linux kernel https://github.com/torvalds/linux/blob/master/include/linux/list.h.
Some small modifications needs to be done to make it work without the other kernel sources  but then it's awesome :-)

waehlisch:
@fho, do you have time to give this PR a review? Would be nice to merge this soon, as this is required to finish the FRR (fork of Quagga) extension.

fho:
@waehlisch Yes, I'll do a review latest on the weekend

rtrlib/rtr_mgr.h
```diff
@@ -67,28 +68,37 @@ struct rtr_mgr_group {
...
+
 typedef void (*rtr_mgr_status_fp)(const struct rtr_mgr_group *,
 				  enum rtr_mgr_status,
 				  const struct rtr_socket *,
 				  void *);
 
+//TODO Add refresh, expire, and retry intervals to config for easier access.
 struct rtr_mgr_config {
-	struct rtr_mgr_group *groups;
+	tommy_list groups;
```
fho: It would be great to hide the actual list implementation from the user. This would allow to change freely the data structure in the background without breaking the API.
I also don't like the idea much that the user has additional to know the tommy_list API to use the rtrlib.

The `struct rtr_mgr_group` should be a opaque data structure that is only accessible via helper functions, something like:
```
struct rtr_mgr_group
struct rtr_mgr_group *rtr_group_alloc();
void rtr_group_free(struct rtr_mgr_group *);
int rtr_mgr_group_append(struct rtr_mgr_group *list, struct rtr_mgr_group *new);
void rtr_mgr_group_remove(struct rtr_mgr_group *grp);


void rtr_mgr_group_add_socket(struct rtr_mgr group *, struct rtr_socket *);
void rtr_mgr_group_set_pref(struct rtr_mgr group *, uint8_t pref);
...
```
Refactoring that API might be a separate topic. The changes in the this pull request are consistent with the previous API but it makes it even more complex :-)

rtrlib/rtr_mgr.c
```diff
@@ -58,21 +59,24 @@ static int rtr_mgr_start_sockets(struct rtr_mgr_group *group)
 	return RTR_SUCCESS;
 }
 
-int rtr_mgr_find_group(const struct rtr_mgr_config *config,
-		       const struct rtr_socket *sock,
-		       unsigned int *ind)
+static struct rtr_mgr_group *rtr_mgr_find_group(struct rtr_mgr_config *config,
+						const struct rtr_socket *sock)
```
fho: 1 space for indentation missing :-)

rtrlib/rtr_mgr.c
```diff
@@ -58,21 +59,24 @@ static int rtr_mgr_start_sockets(struct rtr_mgr_group *group)
...
-			}
+	tommy_node *node;
+	struct rtr_mgr_group_node *group_node;
+
+	node  = tommy_list_head(&config->groups);
+
+	while (node) {
+		group_node = node->data;
+		for (unsigned int j = 0; j < group_node->group->sockets_len;
+									 j++) {
```
fho: why is `j++` on a separate line?

rtrlib/rtr_mgr.c
```diff
@@ -58,21 +59,24 @@ static int rtr_mgr_start_sockets(struct rtr_mgr_group *group)
...
-				return RTR_SUCCESS;
-			}
+	tommy_node *node;
+	struct rtr_mgr_group_node *group_node;
+
+	node  = tommy_list_head(&config->groups);
+
+	while (node) {
+		group_node = node->data;
+		for (unsigned int j = 0; j < group_node->group->sockets_len;
```
fho: `int j` is sufficient and shorter

rtrlib/rtr_mgr.c
```diff
@@ -431,17 +507,148 @@ inline int rtr_mgr_get_spki(struct rtr_mgr_config *config,
...
+	}
+}
+
+int rtr_mgr_add_group(struct rtr_mgr_config *config,
+		      const struct rtr_mgr_group *group)
+{
+	unsigned int refresh_iv = 3600;
+	unsigned int retry_iv = 600;
+	unsigned int expire_iv = 7200;
+	int err_code = RTR_ERROR;
```
fho: instead of int this could be the corresponding enum type

tests/test_live_validation.c
```diff
@@ -116,7 +117,7 @@ int main(void)
 
 	rtr_mgr_stop(conf);
 	rtr_mgr_free(conf);
-	free(groups[0].sockets);
+	//free(groups[0].sockets);
```
fho: either remove the comment or uncomment the line

rtrlib/rtr_mgr.c
```diff
@@ -431,17 +507,148 @@ inline int rtr_mgr_get_spki(struct rtr_mgr_config *config,
...
+
+	new_group_node->group = new_group;
+	tommy_list_insert_tail(&config->groups, &new_group_node->node,
+			       new_group_node);
+	config->len++;
+
+	tommy_list_sort(&config->groups, &rtr_mgr_config_cmp);
+	pthread_mutex_unlock(&config->mutex);
+
+	struct rtr_mgr_group *best_group = rtr_mgr_get_first_group(config);
```
nbars: This pointer might not point to a valid memory region (in the following instructions), as we hold no lock and the group could be freed in the meantime.

rtrlib/rtr_mgr.c
```diff
@@ -90,135 +94,155 @@ bool rtr_mgr_config_status_is_synced(const struct rtr_mgr_group *group)
...
-	/* groups are sorted by preference */
-	for (int i = 0; i < config->len; i++) {
-		if ((i != my_group_idx) &&
-		    (config->groups[i].status == RTR_MGR_CLOSED))
-			return &config->groups[i];
+	tommy_node *node = tommy_list_head(&config->groups);
+	struct rtr_mgr_group_node *group_node;
+	struct rtr_mgr_group *group;
+
+	while (node) {
```
nbars: indention is missing here :-)

rtrlib/rtr_mgr.c
```diff
@@ -431,17 +507,148 @@ inline int rtr_mgr_get_spki(struct rtr_mgr_config *config,
...
 void rtr_mgr_stop(struct rtr_mgr_config *config)
 {
+	tommy_node *node = tommy_list_head(&config->groups);
+	struct rtr_mgr_group_node *group_node;
+
 	MGR_DBG1("rtr_mgr_stop()");
-	for (unsigned int i = 0; i < config->len; i++) {
-		for (unsigned int j = 0; j < config->groups[i].sockets_len; j++)
-			rtr_stop(config->groups[i].sockets[j]);
+	while (node) {
```
nbars: No lock held while iterating over the list.

rtrlib/rtr_mgr.c
ColinBS: In #98 there are default values for these intervals. Could they serve the purpose for this problem?

```
static const uint16_t RTR_EXPIRATION_DEFAULT = 7200;
// ...
static const uint16_t RTR_REFRESH_DEFAULT = 3600;
// ...
static const uint16_t RTR_RETRY_DEFAULT = 600;
```

rtrlib/rtr_mgr.c
```diff
@@ -431,17 +507,148 @@ inline int rtr_mgr_get_spki(struct rtr_mgr_config *config,
...
+
+	struct rtr_mgr_group_node *gnode;
+	tommy_node *node = tommy_list_head(&config->groups);
+
+	while (node) {
+		gnode = node->data;
+		if (gnode->group->preference == group->preference) {
+			RTR_DBG("Group with preference value already exists!");
+			return RTR_INVALID_PARAM;
+		}
```
ColinBS: Could there be a missing unlock before returning?

rtrlib/rtr_mgr.c
mroethke: sockets_len is an unsigned int and so should j

rtrlib/rtr_mgr.c
reuteran: Weird, this looks fine in my editor and also on github when I view the whole file. But in this view there is indeed a space missing...

rtrlib/rtr_mgr.c
reuteran: Coding style script throws a warning because of 80 char line limit

rtrlib/rtr_mgr.c
```diff
@@ -431,17 +525,156 @@ inline int rtr_mgr_get_spki(struct rtr_mgr_config *config,
...
+		//TODO This is not pretty. It wants to get the same intervals
+		//that are being used by other groups. Maybe intervals should
+		//be store globally/per-group/per-socket?
+		if (gnode->group->sockets[0]->refresh_interval)
+			refresh_iv = gnode->group->sockets[0]->refresh_interval;
+		if (gnode->group->sockets[0]->retry_interval)
+			retry_iv = gnode->group->sockets[0]->retry_interval;
+		if (gnode->group->sockets[0]->expire_interval)
+			expire_iv = gnode->group->sockets[0]->expire_interval;
+		node = node->next;
```
ColinBS: I kind of forgot why we are assigning the intervals in the first place here (even though I think I wrote this).
Wouldn't it be better to leave the intervals at their defaults and add API functionality to set intervals? For now, this might only be for the whole group, e.g.:
`set_refresh_interval(struct rtr_mgr_group *group, unsigned int new_interval)`

rtrlib/rtr_mgr.c
```diff
@@ -431,17 +525,156 @@ inline int rtr_mgr_get_spki(struct rtr_mgr_config *config,
...
+			       new_group_node);
+	config->len++;
+
+	tommy_list_sort(&config->groups, &rtr_mgr_config_cmp);
+
+	struct rtr_mgr_group *best_group = rtr_mgr_get_first_group(config);
+
+	if (best_group->status == RTR_MGR_CLOSED) {
+		for (unsigned int j = 0; j < best_group->sockets_len; j++)
+			rtr_start(best_group->sockets[j]);
```
ColinBS: What is the behavior in case one or more sockets cannot be started?
Maybe do
```
if (rtr_start(best_group->sockets[j]) != RTR_SUCCESS)
    tr_free(best_group->sockets[j]);
```
or is this not even necessary?

rtrlib/rtr_mgr.c
```diff
@@ -431,17 +525,152 @@ inline int rtr_mgr_get_spki(struct rtr_mgr_config *config,
...
+	tommy_list_insert_tail(&config->groups, &new_group_node->node,
+			       new_group_node);
+	config->len++;
+
+	tommy_list_sort(&config->groups, &rtr_mgr_config_cmp);
+
+	struct rtr_mgr_group *best_group = rtr_mgr_get_first_group(config);
+
+	if (best_group->status == RTR_MGR_CLOSED) {
+		rtr_mgr_start_sockets(best_group);
```
ColinBS: Still not sure what should happen in case a socket is not started properly by `rtr_mgr_start_sockets`.

rtrlib/rtr_mgr.c
fho: ah, we could remove the `unsigned` instead :-)

rtrlib/rtr_mgr.c
reuteran: Yes, it would be better to be able to set intervals for each socket, since they are already stored by each socket. I think this should be in a separate pull request though.

waehlisch:
To sum up, we have still two open questions:
 1.  How to handle socket specific timers? Suggestion by @reuteran is to put this in a separate PR.
 1. Which implementation for the link list? Suggestion is to postpone. We can beautify this later, it will introduce additionally required features.

@fho, can you give a final review?  We would like to get this merged soon, to progress with FRR ;).

tests/test_dynamic_groups.c
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	//remove 100 groups
+	for (int i = 0; i < 100; i++)
+		rtr_mgr_remove_group(conf, --group3.preference);
+
+	rtr_mgr_stop(conf);
+	rtr_mgr_free(conf);
+	free(group2.sockets);
+	free(groups[0].sockets);
```
nbars: We are missing here a ```free(group3.sockets)``` I think.

rtrlib/rtr_mgr.c
```diff
@@ -331,35 +379,41 @@ int rtr_mgr_init(struct rtr_mgr_config **config_out,
...
-			MGR_DBG1("Error Empty sockets array in socket group!");
-			goto err;
-		}
-		for (unsigned int j = 0; j < cg.sockets_len; j++) {
-			if (rtr_init(cg.sockets[j], NULL,
+	/* Copy the groups from the array into linked list config->groups */
+	config->len = groups_len;
+	config->groups = NULL;
+	struct rtr_mgr_group *cg;
+	struct rtr_mgr_group_node *group_node;
```
fho: variable declaration should only be done at the beginning of the block

rtrlib/rtr_mgr.c
```diff
@@ -331,35 +379,41 @@ int rtr_mgr_init(struct rtr_mgr_config **config_out,
...
+
+	for (unsigned int i = 0; i < groups_len; i++) {
+		cg = malloc(sizeof(struct rtr_mgr_group));
```
fho: check missing if malloc failed

rtrlib/rtr_mgr.c
```diff
@@ -331,35 +379,41 @@ int rtr_mgr_init(struct rtr_mgr_config **config_out,
...
-			MGR_DBG1("Error Same preference for 2 socket groups!");
-			goto err;
-		}
-		if (cg.sockets_len == 0) {
-			MGR_DBG1("Error Empty sockets array in socket group!");
-			goto err;
-		}
-		for (unsigned int j = 0; j < cg.sockets_len; j++) {
-			if (rtr_init(cg.sockets[j], NULL,
+	/* Copy the groups from the array into linked list config->groups */
```
fho: that part could be moved to a separate function to keep rtr_mgr_init() small and better readable

rtrlib/rtr_mgr.c
```diff
@@ -331,35 +379,41 @@ int rtr_mgr_init(struct rtr_mgr_config **config_out,
...
+		memcpy(cg, &groups[i], sizeof(struct rtr_mgr_group));
+
+		cg->status = RTR_MGR_CLOSED;
+		for (unsigned int j = 0; j < cg->sockets_len; j++) {
+			if (rtr_init(cg->sockets[j], NULL,
 				     pfxt, spki_table, refresh_interval,
 				     expire_interval, retry_interval,
-				     rtr_mgr_cb, config) != RTR_SUCCESS) {
+				     rtr_mgr_cb, config, cg) != RTR_SUCCESS) {
 				err_code = RTR_INVALID_PARAM;
```
fho: instead of assigning err_code to the (currently) only error code that rtr_init() can return it would be better to do something like:
```
err = rtr_init();
if (err)
    goto [..]
```
If rtr_init() would change in the future to also return other errors, it's easy to miss to update this place here to ensure that the user gets the right error code back :-)

rtrlib/rtr_mgr.c
```diff
@@ -331,35 +379,41 @@ int rtr_mgr_init(struct rtr_mgr_config **config_out,
...
 				goto err;
 			}
 		}
-		last_preference = cg.preference;
+		group_node = malloc(sizeof(struct rtr_mgr_group_node));
```
fho: check missing if malloc() failed

rtrlib/rtr_mgr.c
```diff
@@ -331,35 +379,41 @@ int rtr_mgr_init(struct rtr_mgr_config **config_out,
...
+		group_node->group = cg;
+		tommy_list_insert_tail(&config->groups, &group_node->node,
+				       group_node);
 	}
+	/* Our linked list should be sorted already, since the groups array was
+	 * sorted. However, for safety reasons we sort again.
+	 */
+	tommy_list_sort(&config->groups, &rtr_mgr_config_cmp);
```
fho: that's unnecessary, it **must** be sorted already :-)

rtrlib/rtr_mgr.c
fho: The allocated memory seems to be leaked if the function returns because rtr_init() failed

rtrlib/rtr_mgr.c
```diff
@@ -377,37 +431,77 @@ int rtr_mgr_init(struct rtr_mgr_config **config_out,
 	return err_code;
 }
 
+static struct rtr_mgr_group *rtr_mgr_get_first_group(struct rtr_mgr_config
```
fho: name could be shortened to `rtr_mgr_first_group`

rtrlib/rtr_mgr.c
```diff
@@ -377,37 +431,77 @@ int rtr_mgr_init(struct rtr_mgr_config **config_out,
...
+						     *config)
+{
+	tommy_node *head = tommy_list_head(&config->groups);
+	struct rtr_mgr_group_node *group_node = head->data;
```
fho: the assignment could be removed and return head->data->group directly

rtrlib/rtr_mgr.c
```diff
@@ -377,37 +431,77 @@ int rtr_mgr_init(struct rtr_mgr_config **config_out,
 	return err_code;
 }
 
+static struct rtr_mgr_group *rtr_mgr_get_first_group(struct rtr_mgr_config
+						     *config)
+{
+	tommy_node *head = tommy_list_head(&config->groups);
```
fho: the function doesn't need to acquire the mutex?

rtrlib/rtr_mgr.c
```diff
@@ -431,17 +525,155 @@ inline int rtr_mgr_get_spki(struct rtr_mgr_config *config,
...
+int rtr_mgr_add_group(struct rtr_mgr_config *config,
+		      const struct rtr_mgr_group *group)
+{
+	unsigned int refresh_iv = 3600;
+	unsigned int retry_iv = 600;
+	unsigned int expire_iv = 7200;
+
+	pthread_mutex_lock(&config->mutex);
+
+	struct rtr_mgr_group_node *gnode;
```
fho: declaration should be done before statements

rtrlib/rtr_mgr.c
```diff
@@ -431,17 +525,155 @@ inline int rtr_mgr_get_spki(struct rtr_mgr_config *config,
...
+	tommy_node *node = tommy_list_head(&config->groups);
+
+	while (node) {
+		gnode = node->data;
+		if (gnode->group->preference == group->preference) {
+			MGR_DBG1("Group with preference value already exists!");
+			pthread_mutex_unlock(&config->mutex);
+			return RTR_INVALID_PARAM;
```
fho: instead of having to enable debugging to find out the  cause it would be great to return an error code that describes the reason more specific

rtrlib/rtr_mgr.c
```diff
@@ -431,17 +525,155 @@ inline int rtr_mgr_get_spki(struct rtr_mgr_config *config,
...
+		}
+	}
+
+	struct rtr_mgr_group_node *new_group_node = malloc(
+					sizeof(struct rtr_mgr_group_node));
+	if (!new_group_node) {
+		pthread_mutex_unlock(&config->mutex);
+		free(new_group);
```
fho: use a goto error handler to prevent duplicating code in case of an error

rtrlib/rtr_mgr.c
```diff
@@ -431,17 +525,155 @@ inline int rtr_mgr_get_spki(struct rtr_mgr_config *config,
...
+	struct rtr_mgr_group *new_group = malloc(sizeof(struct rtr_mgr_group));
+
+	if (!new_group) {
+		pthread_mutex_unlock(&config->mutex);
+		return RTR_ERROR;
+	}
+	memcpy(new_group, group, sizeof(struct rtr_mgr_group));
+	new_group->status = RTR_MGR_CLOSED;
+
+	for (unsigned int j = 0; j < new_group->sockets_len; j++) {
```
fho: maybe that can be outsourced to a separate function and the same code can be used from rtr_mgr_init() and rtr_mgr_add_group()

rtrlib/rtr_mgr.c
```diff
@@ -431,17 +525,155 @@ inline int rtr_mgr_get_spki(struct rtr_mgr_config *config,
...
+int rtr_mgr_remove_group(struct rtr_mgr_config *config,
+			 unsigned int preference)
+{
+	pthread_mutex_lock(&config->mutex);
+	tommy_node *remove_node = NULL;
+	tommy_node *node = tommy_list_head(&config->groups);
+	struct rtr_mgr_group_node *group_node;
+	struct rtr_mgr_group *remove_group;
+
+	// Find the node of the group we want to remove
```
fho: could be a separate function _find_node() or _remove_node() (which also finds it)

rtrlib/rtr_mgr.h
```diff
@@ -67,28 +68,37 @@ struct rtr_mgr_group {
 	enum rtr_mgr_status status;
 };
 
+//TODO Find a nicer way todo a linked list (without writing our own)
+struct rtr_mgr_group_node {
```
fho: Seems that the `struct rtr_mgr_group_node` is only used internally in rtr_mgr.c if it's true it should be moved to the .c file

fho:
> To sum up, we have still two open questions:
> 
> How to handle socket specific timers? Suggestion by @reuteran is to put this in a separate PR.

sounds good to me

> Which implementation for the link list? Suggestion is to postpone. We can beautify this later, it will introduce additionally required features.

Seems that the list is only used internally on contrary what I first thought. :-)
I agree, which linked list implementation is used doesn't really matter here.

> @fho, can you give a final review? We would like to get this merged soon, to progress with FRR ;).

Yes, sorry, I was spontaneously on a weekend trip and didn't manage to do it like planned.
I added more comments but I still wasn't able to go through all of the code neither to understand all the logic.
I'll try to have another look soon, if it wasn't merged then already. :-)
It looks good in general, cool improvement!

rtrlib/rtr_mgr.h
ColinBS: The struct is also used in the `test_dynamic_groups` test. So we keep it in the header file?

rtrlib/rtr_mgr.c
ColinBS: For some reason this won't work. No idea why not, though...

rtrlib/rtr_mgr.c
ColinBS: I will try to make this a PR on it's own. @mroethke suggested to use the `add_group` function but there might be some refactoring necessary to make this happen.

rtrlib/rtr_mgr.c
ColinBS: This might also be changed in a separate PR that does some refactoring. In my humble opinion, I think the name is OK, the _get_ prevents confusion over what exactly the function does, I think.

rtrlib/rtr_mgr.c
ColinBS: Would like to ask @reuteran and @mroethke about this.

rtrlib/rtr_mgr.c
ColinBS: Not sure. @reuteran and @mroethke ?

rtrlib/rtr_mgr.c
ColinBS: Could be part of a separate PR.

rtrlib/rtr_mgr.c
fho: I meant the `err_code` variable :

rtrlib/rtr_mgr.h
fho: No, the external API should be minimal and only contain declaration that have a purpose during normal use of the library.
Either the test should use only the external library or we have to find another way to test it. (including the .c file in the test might work)

mroethke: We need a way to access the the group list tough or at least access the active group, otherwise it would not be possible to get status information beyond "it still works somehow".

Of course that can also be archived with helper functions.

rtrlib/rtr_mgr.c
ColinBS: Ohhhh... well... huh...


Will do that!

mroethke:
I just re based this onto master and made the necessary changes to properly use the alloc utils. 

I furthermore made a change to how socket initialization handles the configuration. While not strictly related to dynamic groups it fixes an error that could could lead to memory leaks that is basically irrelevant without this patch.

rtrlib/rtr_mgr.c
smlng: why isn't `sockets_len` of type `size_t` anyway? Or alternatively just `unsigned` should suffice, too.

rtrlib/rtr_mgr.c
```diff
@@ -91,135 +120,171 @@ bool rtr_mgr_config_status_is_synced(const struct rtr_mgr_group *group)
...
-		    (config->groups[i].status == RTR_MGR_CLOSED))
-			return &config->groups[i];
+	struct rtr_mgr_group_node *group_node;
+	struct rtr_mgr_group *group;
+
+	pthread_mutex_lock(&config->mutex);
+	tommy_node *node = tommy_list_head(&config->groups);
+
+	while (node) {
+	group_node = node->data;
```
smlng: indention looks of by one tab here

rtrlib/rtr_mgr.c
```diff
@@ -91,135 +120,171 @@ bool rtr_mgr_config_status_is_synced(const struct rtr_mgr_group *group)
...
 static bool is_some_rtr_mgr_group_established(struct rtr_mgr_config *config)
 {
-	for (int i = 0; i < config->len; i++) {
-		if (config->groups[i].status == RTR_MGR_ESTABLISHED)
+	struct rtr_mgr_group_node *group_node;
+
+	pthread_mutex_lock(&config->mutex);
+	tommy_node *node = tommy_list_head(&config->groups);
+
+	while (node) {
```
smlng: a for-loop would fit here perfectly, or not?

```
for (tommy_node *node = tommy_list_head(&config->groups); 
     node != NULL; node = node->next) {
...
}
```

rtrlib/rtr_mgr.c
```diff
@@ -91,135 +120,171 @@ bool rtr_mgr_config_status_is_synced(const struct rtr_mgr_group *group)
...
-		set_status(config, &config->groups[ind],
-			   config->groups[ind].status, sock);
+		set_status(config, group,
+			   group->status, sock);
 }
 
 static inline void _rtr_mgr_cb_state_established(const struct rtr_socket *sock,
 						 struct rtr_mgr_config *config,
-						 unsigned int ind)
+						 struct rtr_mgr_group *my_group)
```
smlng: inconsistent parameter naming, above its just `group` which is fine - `my_group` looks a bit like code from a beginners tutorial 😉

rtrlib/rtr_mgr.c
```diff
@@ -91,135 +120,171 @@ bool rtr_mgr_config_status_is_synced(const struct rtr_mgr_group *group)
...
-			    cg.status != RTR_MGR_ERROR &&
-			    cg.status != RTR_MGR_CLOSED &&
-			    cg.preference < config->groups[ind].preference)
+		struct rtr_mgr_group_node *group_node;
+		struct rtr_mgr_group *current_group;
+
+		pthread_mutex_lock(&config->mutex);
+		tommy_node *node = tommy_list_head(&config->groups);
+
+		while (node) {
```
smlng: again for-loop might fit better here, see above

rtrlib/rtr_mgr.c
```diff
@@ -91,135 +120,171 @@ bool rtr_mgr_config_status_is_synced(const struct rtr_mgr_group *group)
...
+			group_node = node->data;
+			current_group = group_node->group;
+			if (current_group != my_group &&
+			    current_group->status != RTR_MGR_ERROR &&
```
smlng: if would put parentheses around each condition

rtrlib/rtr_mgr.c
```diff
@@ -371,44 +449,87 @@ int rtr_mgr_init(struct rtr_mgr_config **config_out,
...
+	bool all_sync;
+	struct rtr_mgr_group_node *group_node;
 
-		for (unsigned int j = 0;
-		     all_sync && (j < config->groups[i].sockets_len); j++) {
-			if (config->groups[i].sockets[j]->last_update == 0)
+	pthread_mutex_lock(&config->mutex);
+	tommy_node *node = tommy_list_head(&config->groups);
+
+	while (node) {
```
smlng: use for-loop? See previous comments ...

rtrlib/rtr_mgr.c
colinbs: In my opinion, the readability suffers from this more compact solution.

rtrlib/lib/alloc_utils.h
```diff
@@ -34,4 +34,13 @@ void lrtr_free(void *ptr);
 
 void *lrtr_realloc(void *ptr, size_t size);
 
+/**
+ * @brief Duplicates a string
+ * @param[in] string
+ * @returns Duplicated string
+ * @returns NULL if string is NULL
+ * @returns NULL on error
```
mroethke: The main and currently only use case for this function is copying the socket configuration, parts of which may be null. Instead of guarding every copy I decided to write the function accordingly.

mroethke: You are right.
I changed to function to match std strdup and added checks where necessary.

rtrlib/rtr_mgr.c
smlng: fine by me, leave as is.

smlng:
@mroethke and @colinbs please consider #132

PR #135, opened 2017-09-26
Use static memory for config, pfx_table, and spki_table

smlng:
This PR simplifies memory handling for config, pfx_table, and spki_table by using static (stack) memory instead of dynamic mem with malloc. Further, it ensures that only a single RTR manager can run at a time.

### DIFF
--- a/rtrlib/rtr_mgr.c
+++ b/rtrlib/rtr_mgr.c
@@ -27,6 +27,10 @@ static const char * const mgr_str_status[] = {
 	[RTR_MGR_ERROR] = "RTR_MGR_ERROR",
 };
 
+static struct pfx_table rtr_pfx_table;
+static struct spki_table rtr_spki_table;
+static struct rtr_mgr_config rtr_config;
+
 static struct rtr_mgr_group *rtr_mgr_find_group(struct rtr_mgr_config *config,
 						const struct rtr_socket *sock);
 static int rtr_mgr_config_cmp(const void *a, const void *b);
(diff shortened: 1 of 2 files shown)


rtrlib/rtr_mgr.c
```diff
@@ -721,13 +711,13 @@ int rtr_mgr_remove_group(struct rtr_mgr_config *config,
 }
 
 // TODO: write test for this function.
-int rtr_mgr_for_each_group(struct rtr_mgr_config *conf,
+int rtr_mgr_for_each_group(struct rtr_mgr_config *config,
```
mroethke: Please sync signature with declaration in the header.

smlng: ? it matches, besides the function pointer (fp) is used here as with all other `_for_each_` functions, see below. Or what do you mean?

nbars:
So this change makes the ```rtr_mgr``` to something similar to a singleton, right?
This defeats the purpose of the ```struct rtr_mgr_config``` references that are passed around to use the library functions (because there can be only one instance). 

I think this is not really an improvement of the API because the user of the library expects that he can instantiate as many objects of the type ```struct rtr_mgr_config``` as he likes (because of the init function that returns an object of this type). But this is not the case anymore (this should be documented in function header), and in my opinion this is really counter intuitive.

I think there are two options:
1. Update the documentation and remove all reference passing related to the ```rtr_mgr_``` stuff. Also, there must be some ways to set ```rtr_config.running``` back to ```false``` (did I missed that?).

2. (Preferred) Leave the API as it is and postpone theses changes up to the point where the redesign of the API is made. When the API is redesigned, this code will be rewritten anyway.

Beside the points stated above, this change also introduces a higher coupling between the ```rtr_mgt``` and the ```spki``` and ```pfx``` module because you are instantiating the corresponding structures in the ```.bss``` section (static keyword) and this prevents us from hiding the implementation of e.g. ```struct pfx_table``` by moving it into a ```.c``` file (as it should be).

smlng:
closing, should be rethought when API is reworked, too.

PR #137, opened 2017-11-07
rtrlib, version: intermediate update to 0.5.0 for FRR integration

smlng:
This PR increments the version patch part from `0` to `1`, effectively updating the rtrlib version to `0.4.1`. This is required for the upcoming FRR RPKI integration using rtrlib, because latest official release of the rtrlib with version `0.4.0` does not have the dynamic group feature required by FRR. Hence, increasing the version of rtrlib to `0.4.1` ensures that FRRs configure script can check for rtrlib version > `0.4.0` which will fail for the current release (and master) but succeed with this PR (and any future rtrlib release). 

So this is an intermediate patch until we make the next official release, which then might be already version `0.5.0` - which will also match version > `0.4.0` and thus no changes to FRR are needed.


smlng:
updated CHANGELOG

fho:
If I remember it right the rtr_mgr API changed. Shouldn't  we increase the minor number  instead of the patch nr?
Version 0.5?

smlng:
@fho as said above this version bump is intermediate so that @mroethke and @colinbs can move forward with there RTRlib integration in FRR. Meaning, that the FRR configure script can assert that rtrlib with version > `0.4.0` is installed in the system. Which would thus fail for the last release but succeed for RTRlib master with this PR merged. In that sense, I would not make `0.4.1` an official release but rather have it here on GitHub to work with FRR until we make the next official RTRlib release, which certainly could be `0.5.0` <- though I do not see how the API changed until now.

smlng:
> though I do not see how the API changed until now.

IIRC at least no existing call was changed, only new calls for add/del of groups.

waehlisch:
I don't see a reason why we shouldn't publish a new release. Dynamic group management is a new feature.

smlng:
> I don't see a reason why we shouldn't publish a new release.

My initial intention with this PR was to not make much noise about it and just have things in place for the FRR integration to progress smoothly.

As this didn't work out, we can now proceed and make a new release: which version number is it going to be `0.4.1` or `0.5.0` -  I think the latter is fine, as we have new features and not only bugfixes?

waehlisch:
OK

smlng:
changed version to `0.5.0`, @waehlisch what else needs to be done for a release?

Btw. needs squashing now

waehlisch:
@smlng, just address my comment about the order of changes in the CHANGELOG.
You also need to update debian/changelog

smlng:
>  just address my comment about the order of changes 

??? which one, I guess you mean to have the feature (dynamic groups) named first?

waehlisch:
yes, first mention the introduction of a new feature and then tests

PR #139, opened 2017-11-20
Install .pc via librtr-dev package

xdel:


waehlisch:
Thanks @xdel!

mroethke:
Thank you very much. I stumbled upon this myself recently, but did not have time to fix this yet.

@waehlisch 
I suggest we push this through quickly and release it as bug fix.

PR #154, opened 2018-02-08
trie: fix deletion of non leaf nodes

### LINKED ISSUE #152: Errors when withdrawing previously added VRP
While testing the new rpki-validator v3, people at RIPE found a problem with the RTRlib:
> We found one issue testing with RTRLib that we believe is not us: we got errors from RTRLib when we withdraw VRPs - even though the same VRP was just reported as added.

The rpki-rtr server they are using is here (rpm package, use Centos7):
https://github.com/RIPE-NCC/rpki-validator-3/wiki/RIPE-NCC-RPKI-Validator-3-beta-tester-page

Details about feeding test data to that rpki-rtr server, to reproduce the RTRlib error:

> You can change properties for the validator and rtr server in these files (restart required):
/etc/rpki-validator-3/application.properties
/etc/rpki-rtr-server/application.properties
The best way to test the RTR server may be to modify the ‘rpki.validator.validated.roas.uri’ property to point to somewhere other than the actual validator, and put some (small) json file there that follows the same format. I have not tried it, but it may work if you use a ‘file:///' url.

mroethke:
When a non leaf node is deleted one if its childs will takes its place.
The selected child must be the one with the shorter prefix length.
If the one with the longer prefix length is selected the node with the
shorter prefix length will be buried, meaning the search algorithm will
not find it anymore.

Fixes #152

### DIFF
--- a/tests/test_pfx.c
+++ b/tests/test_pfx.c
@@ -330,6 +330,16 @@ static void pfx_table_test(void)
 	printf("pfx_table_test successful\n");
 }
 
+static void create_ip4_pfx_record(struct pfx_record *pfx, uint32_t asn,
+				  const char *ip, uint8_t min_mask_len,
+				  uint8_t max_mask_len)
+{
+	pfx->asn = asn;
+	pfx->min_len = min_mask_len;
+	pfx->max_len = max_mask_len;
+	assert(!lrtr_ip_str_to_addr(ip, &pfx->prefix));
+}
+
 static void add_ip4_pfx_record(struct pfx_table *pfxt, uint32_t asn,
 			       const char *ip, uint8_t min_mask_len,
 			       uint8_t max_mask_len)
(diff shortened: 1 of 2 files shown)


tests/test_pfx.c
```diff
@@ -330,18 +330,25 @@ static void pfx_table_test(void)
 	printf("pfx_table_test successful\n");
 }
 
+static void create_ip4_pfx_record(struct pfx_record *pfx, uint32_t asn,
+			       const char *ip, uint8_t min_mask_len,
```
smlng: indention looks off? not aligned with parentheses above?

tests/test_pfx.c
```diff
@@ -363,12 +370,37 @@ static void test_issue99(void)
...
+	create_ip4_pfx_record(&records[5], 6, "213.175.86.0", 24, 24);
+
+	for (size_t i = 0; i < 6; i++) {
+		assert(pfx_table_add(&pfxt, &records[i]) == PFX_SUCCESS);
+	}
+
+	for (size_t i = 0; i < 6; i++) {
+		assert(pfx_table_remove(&pfxt, &records[i]) == PFX_SUCCESS);
+	}
+
```
smlng: remove empty lines

tests/test_pfx.c
```diff
@@ -363,12 +370,37 @@ static void test_issue99(void)
 	pfx_table_free(&pfxt);
 }
 
+static void test_issue152(void)
+{
+	struct pfx_table pfxt;
+	struct pfx_record *records = calloc(6, sizeof(struct pfx_record));
```
smlng: unsure myself, but why not use `struct pfx_record records[6];` instead?

tests/test_pfx.c
smlng: but maybe my browser is to blame here ...

reuteran: indent is off on my screen too

tests/test_pfx.c
reuteran: free() statement missing for 'records'?

tests/test_pfx.c
mroethke: No specific reason, I want initialized memory there and apparently calloc occurred to me before memset did.

smlng:
@reuteran are we good here?

PR #155, opened 2018-02-08
scripts: fix travis.sh

mroethke:
Due to an error in the travis script the travis build can not fail
because the return value of failed build steps are never propagated to
the build engine.

This commit is expected to fail travis.


smlng:
can you please provide a fix for cppcheck issues, thx!

Issue #159, opened 2018-03-15
rtrlib dev packages should require libssh dev packages


mwinter-osr:
Building a package with librtr dev package should require libssh dev package

I assume the librtr.pc file is correct, which contains (on Debian):
```
prefix=/usr
includedir=${prefix}/include
libdir=${prefix}/lib/x86_64-linux-gnu

Name: rtrlib
Description: 
Version: 0.5.0
Requires: libssh >= 0.5.0
Libs: -L${libdir} -lrtr
Cflags: -I${includedir}
```

Based on this file, it requires libssh and building a package which requires the librtr-dev
(ie FRR), causes a failure on `configure`:
```
configure:25768: checking for RTRLIB
configure:25775: $PKG_CONFIG --exists --print-errors "rtrlib >= 0.5.0"
Package libssh was not found in the pkg-config search path.
Perhaps you should add the directory containing `libssh.pc'
to the PKG_CONFIG_PATH environment variable
Package 'libssh', required by 'rtrlib', not found
```

(`libssh.pc` is part of the libssh-dev package)

waehlisch:
Fixed in #160.

PR #160, opened 2018-03-18
redhat/debian: Dev package needs to require libssh dev package

mwinter-osr:
This fixes issue #159 

Signed-off-by: Martin Winter <mwinter@opensourcerouting.org>

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -25,7 +25,7 @@ Section: libdevel
 Architecture: any
 Multi-Arch: same
 Pre-Depends: ${misc:Pre-Depends}
-Depends: librtr0 (= ${binary:Version}), ${misc:Depends}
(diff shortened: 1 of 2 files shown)


mroethke:
debian has it already specified as build dependency (in line 5 of the control file) which should be enough.
Should it not be a build time dependency for redhat as well?

mwinter-osr:
Yes, it already was specified as `build dependency`. But these are only dependencies required to built the rtrlib package itself.
To USE the rtrlib-dev package (only dev package) and build your own software depending on it, it depends on the libssh dev package as well. So what this issue and PR does is adding it as a normal dependency to the dev package.

I ran into this issue when I've built FRR on a system which didn't have libssh-dev installed. The standard configure script fails the check for librtr as the `librtr.pc` file refers to `libssh.pc` which was missing.

colinbs:
Ah ok I see. This sounds very convenient to me.

PR #162, opened 2018-04-18
cmake: fix a bug that prevents rtrlib from using libssh versions above 0.5.0.

colinbs:
There is a bug that always downgrades libssh usage to 0.5.0. This one does not affect SSH functionality at all, since libssh >0.5.0 has legacy support for 0.5.0.

The bug was caused by the CMake variable `LIBSSH_060` not being visible inside the `ssh_transport.c` file. Necessary variables are now added to the `CMAKE_C_FLAGS`.

RTRlib also no longer checks for specific libssh versions. It takes any that is higher than 0.5.0.


smlng:
@mroethke any objections?

colinbs:
Oh and while we are at it: `LIBSSH_LIBRARY` and `LIBSSH_INCLUDE` which are described in the README are not present in the cmake file. There, they are called `LIBSSH_LIBRAR*IES*` and `LIBSSH_INCLUDE*_DIRS*`. Or maybe fix this in a separate PR?

PR #165, opened 2018-05-03
Implement atomic reset

mroethke:
If a cache server is unable to perform a incremental update rtrlib resets the socket and deletes all records from that server. Once this is done it starts to recieve new records and then adds them to the tables.
In the time period between the reset and the completion of the add operation the state of the spki and pfx tabe is invalid. Validations in that time period will likely have wrong results.

This implements an atomic reset using a shadow table. It incurs a notable delay in the operations of rtrlibs thread but does not affect validation. Since a cache reset should be rare with properly implemented cache servers this should not be a problem in real world scenarios.

### DIFF
--- a/rtrlib/pfx/pfx.h
+++ b/rtrlib/pfx/pfx.h
@@ -189,7 +189,7 @@ void pfx_table_for_each_ipv6_record(struct pfx_table *pfx_table, pfx_for_each_fp
  * @param[out] dst_table Destination table
  * @param[in] socket socket which prefixes should not be copied
  */
-void pfx_table_copy_except_socket(struct pfx_table *src_table, struct pfx_table *dst_table, const struct rtr_socket *socket);
+int pfx_table_copy_except_socket(struct pfx_table *src_table, struct pfx_table *dst_table, const struct rtr_socket *socket);
 
 /**
  * @brief Swap root nodes of the argument tables
--- a/rtrlib/pfx/pfx.h
+++ b/rtrlib/pfx/pfx.h
@@ -183,18 +183,18 @@ void pfx_table_for_each_ipv4_record(struct pfx_table *pfx_table, pfx_for_each_fp
 void pfx_table_for_each_ipv6_record(struct pfx_table *pfx_table, pfx_for_each_fp fp, void *data);
 
 /**
- * @brief Copy content of src into dst
+ * @brief Copy content of @p src_table into @p dst_table
  * @details dst must be empty and initialized
  * @param[in] src_table Source table
- * @param[in] dst_table Destination table
+ * @param[out] dst_table Destination table
  * @param[in] socket socket which prefixes should not be copied
  */
 void pfx_table_copy_except_socket(struct pfx_table *src_table, struct pfx_table *dst_table, const struct rtr_socket *socket);
 
 /**
  * @brief Swap root nodes of the argument tables
- * @param[in] a First table
- * @param[in] b second table
+ * @param[in,out] a First table
+ * @param[in,out] b second table
  */
 void pfx_table_swap(struct pfx_table *a, struct pfx_table *b);
 
--- a/rtrlib/pfx/pfx.h
+++ b/rtrlib/pfx/pfx.h
@@ -100,6 +100,13 @@ void pfx_table_init(struct pfx_table *pfx_table, pfx_update_fp update_fp);
  */
 void pfx_table_free(struct pfx_table *pfx_table);
 
+
+/**
+ * @brief Frees all memory associcated with the pfx_table without calling the update callback.
+ * @param[in] pfx_table pfx_table that will be freed.
+ */
+void pfx_table_free_without_notify(struct pfx_table *pfx_table);
+
 /**
  * @brief Adds a pfx_record to a pfx_table.
  * @param[in] pfx_table pfx_table to use.
@@ -174,6 +181,31 @@ void pfx_table_for_each_ipv4_record(struct pfx_table *pfx_table, pfx_for_each_fp
  * @param[in] data This parameter is forwarded to the callback function.
  */
 void pfx_table_for_each_ipv6_record(struct pfx_table *pfx_table, pfx_for_each_fp fp, void *data);
+
+/**
+ * @brief Copy content of src into dst
+ * @details dst must be empty and initialized
+ * @param[in] src_table Source table
+ * @param[in] dst_table Destination table
+ * @param[in] socket socket which prefixes should not be copied
+ */
+void pfx_table_copy_except_socket(struct pfx_table *src_table, struct pfx_table *dst_table, const struct rtr_socket *socket);
+
+/**
+ * @brief Swap root nodes of the argument tables
+ * @param[in] a First table
+ * @param[in] b second table
+ */
+void pfx_table_swap(struct pfx_table *a, struct pfx_table *b);
+
+/**
+ * @brief Notify client about changes between to pfx tables regarding one specific socket
+ * @details old_table will be modified it should be freed after calling this function
+ * @param[in] new_table
+ * @param[in] old_table
+ * @param[in] socket socket which prefixes should be diffed
+ */
+void pfx_table_notify_diff(struct pfx_table *new_table, struct pfx_table *old_table, const struct rtr_socket *socket);
 
 #endif
 /* @} */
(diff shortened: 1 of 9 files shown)


rtrlib/rtr/packets.c
```diff
@@ -1080,14 +1083,38 @@ int rtr_sync_receive_and_store_pdus(struct rtr_socket *rtr_socket){
...
+                pfx_table_copy_except_socket(rtr_socket->pfx_table, pfx_update_table, rtr_socket);
+
+                spki_table_init(&spki_shadow_table, NULL);
+                spki_update_table = &spki_shadow_table;
+                if (spki_table_copy_except_socket(rtr_socket->spki_table, spki_update_table, rtr_socket) != SPKI_SUCCESS) {
+                    RTR_DBG1("Creation of spki shadow table failed");
+                    rtr_change_socket_state(rtr_socket, RTR_ERROR_FATAL);
+                    retval = RTR_ERROR;
+                    goto cleanup;
+                }
```
reuteran: Its a bit annoying that spki and pfx table have different code here for the same thing, but its necessary. Maybe at some point we can rework the spki table to offer a 'for_each' function and re-implement spki_table_copy_except_socket so its similar to pfx_table_copy_except_socket. The tommy library might actually offer something like this already ( a for-each function)

rtrlib/spki/hashtable/ht-spkitable.c
```diff
@@ -288,3 +288,89 @@ int spki_table_src_remove(struct spki_table *spki_table,
...
+    pthread_rwlock_rdlock(&src->lock);
+    current_node = tommy_list_head(&src->list);
+    while (current_node) {
+        struct key_entry *entry;
+        struct spki_record record;
+
+        entry = (struct key_entry *) current_node->data;
+        key_entry_to_spki_record(entry, &record);
+
+        if (entry->socket == socket) {
```
reuteran: Shouldn't this be entry->socket != socket ?

tests/test_ht_spkitable.c
```diff
@@ -538,6 +538,156 @@ static void test_ht_7(void)
...
+	assert(result_len == 1);
+	free(result);
+	result = NULL;
+
+	assert(spki_table_search_by_ski(
+				&table2,
+				test_record2->ski,
+				&result,
+				&result_len)
+			== SPKI_SUCCESS);
```
reuteran: Weird formatting

rtrlib/pfx/trie/trie-pfx.c
```diff
@@ -505,3 +511,89 @@ void pfx_table_for_each_ipv6_record(struct pfx_table *pfx_table, pfx_for_each_fp
     pfx_table_for_each_rec(pfx_table->ipv6, fp, data);
     pthread_rwlock_unlock(&pfx_table->lock);
 }
+
+struct copy_cb_args {
+	struct pfx_table *pfx_table;
+	const struct rtr_socket *socket;
+};
```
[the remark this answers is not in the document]
mroethke: Our public api surface is already big enough no need to add internal structs to that, especially when there are only used in one file.
However they probably should be declared static and (as suggested by @nbars) moved to the top of the file.

rtrlib/spki/hashtable/ht-spkitable.c
mroethke: Yes, it should …

tests/test_ht_spkitable.c
mroethke: how about 
```
        assert(spki_table_search_by_ski(&table2, test_record2->ski, &result,
                                        &result_len)
               == SPKI_SUCCESS);
```
?

rtrlib/pfx/pfx.h
```diff
@@ -175,5 +182,30 @@ void pfx_table_for_each_ipv4_record(struct pfx_table *pfx_table, pfx_for_each_fp
  */
 void pfx_table_for_each_ipv6_record(struct pfx_table *pfx_table, pfx_for_each_fp fp, void *data);
 
+/**
+ * @brief Copy content of src into dst
```
smlng: nit picky, its not src and dst, should be `Copy content of @p src_table into @p dst_table`

rtrlib/pfx/pfx.h
```diff
@@ -175,5 +182,30 @@ void pfx_table_for_each_ipv4_record(struct pfx_table *pfx_table, pfx_for_each_fp
...
+ * @details dst must be empty and initialized
+ * @param[in] src_table Source table
+ * @param[in] dst_table Destination table
+ * @param[in] socket socket which prefixes should not be copied
+ */
+void pfx_table_copy_except_socket(struct pfx_table *src_table, struct pfx_table *dst_table, const struct rtr_socket *socket);
```
smlng: src_table could be const, too?

rtrlib/pfx/pfx.h
```diff
@@ -175,5 +182,30 @@ void pfx_table_for_each_ipv4_record(struct pfx_table *pfx_table, pfx_for_each_fp
  */
 void pfx_table_for_each_ipv6_record(struct pfx_table *pfx_table, pfx_for_each_fp fp, void *data);
 
+/**
+ * @brief Copy content of src into dst
+ * @details dst must be empty and initialized
+ * @param[in] src_table Source table
+ * @param[in] dst_table Destination table
```
smlng: that's actually an `out` param, or not?

rtrlib/pfx/pfx.h
```diff
@@ -175,5 +182,30 @@ void pfx_table_for_each_ipv4_record(struct pfx_table *pfx_table, pfx_for_each_fp
...
+ * @param[in] socket socket which prefixes should not be copied
+ */
+void pfx_table_copy_except_socket(struct pfx_table *src_table, struct pfx_table *dst_table, const struct rtr_socket *socket);
+
+/**
+ * @brief Swap root nodes of the argument tables
+ * @param[in] a First table
```
smlng: mhm I think `a` and `b`  are `in,out` params not only `in`

rtrlib/pfx/pfx.h
```diff
@@ -175,5 +182,30 @@ void pfx_table_for_each_ipv4_record(struct pfx_table *pfx_table, pfx_for_each_fp
...
+void pfx_table_swap(struct pfx_table *a, struct pfx_table *b);
+
+/**
+ * @brief Notify client about changes between to pfx tables regarding one specific socket
+ * @details old_table will be modified it should be freed after calling this function
+ * @param[in] new_table
+ * @param[in] old_table
+ * @param[in] socket socket which prefixes should be diffed
+ */
+void pfx_table_notify_diff(struct pfx_table *new_table, struct pfx_table *old_table, const struct rtr_socket *socket);
```
smlng: can't `new_table` be const here?

rtrlib/spki/spkitable.h
```diff
@@ -141,5 +141,29 @@ int spki_table_remove_entry(struct spki_table *spki_table,
 int spki_table_src_remove(struct spki_table *spki_table,
 			  const struct rtr_socket *socket);
 
+/**
+ * @brief Copy spki table except entries from the given socket
+ * @param[in] src source table
+ * @param[in] dest target table
+ * @param[in] rtr_socket socket which entries should not be copied
```
smlng: `@return` missing

rtrlib/spki/hashtable/ht-spkitable.c
smlng: how about fix it 😄

rtrlib/pfx/pfx.h
mroethke: Unfortunately no, this function calls `pfx_table_for_each_ipv{4,6}_record` which must mutate the source table to acquire a read lock.

mroethke:
Did you mean `*_free_without_notify`?
If yes you are right, I forgot to implement that for the spki table.

Regarding returning errors. `pfx_table_copy_except_socket` should check of pfx_table_add succeeds and return it's error if not. I think the other functions are fine.

rtrlib/spki/hashtable/ht-spkitable.c
mroethke: whuups I thought I already did …

waehlisch:
Is this PR ready to merge in the next few days? I'm asking because of the next release.

mroethke:
I believe I addressed everything except the formatting thing in `test_ht_spkitable.c` but there was no response to my inquiry.
If I missed anything, please point that out so I can take care of that tomorrow.

@smlng 
I think that is a good idea, but should be it's own PR

tcschmidt:
Ping @reuteran - @mroethke waits for your review :-)

PR #172, opened 2018-08-14
tools: add man pages

### LINKED ISSUE #145: Write man pages for rtrclient and cli-validator
Package linting tools like lintian complain about our tools not having man pages and of course they are right about that.

mroethke:
This kind of depends on #169 because it assumes the name change is already made.

Fix #145

### DIFF
--- a/tools/rpki-rov.1
+++ b/tools/rpki-rov.1
@@ -1,0 +1,29 @@
+.\"
+.\" This file is part of RTRlib.
+.\"
+.\" This file is subject to the terms and conditions of the MIT license.
+.\" See the file LICENSE in the top level directory for more details.
+.\"
+.\" Website: http://rtrlib.realmv6.org/
+.\"
+.TH "rpki-rov" "1"
+.SH NAME
+rpki-rov - rpki route origin validator
+.SH SYNOPSIS
+.B rpki-rov
+.IR HOST
+.IR PORT
+.SH DESCRIPTION
+\fBrpki-rov\fR is a command line interface that implements RPKI route origin validation.
+It connects to the rpki server described by \fIhost\fR and \fIport\fR and can be used interactively or via pipes.
+The format of the input parameter is always "\fIIP PREFIXLENGTH ASN\fR" followed by a newline.
+.SH EXAMPLES
+Pipe into stdin
+.PP
+.nf
+.RS
+echo "93.175.146.0 24 12654" | rpki-rov rpki.example.com 8282
+.RE
+.fi
+.SH BUGS
+Does currently not support ssh-based rpki server.
--- a/tools/rtrclient.1
+++ b/tools/rtrclient.1
@@ -1,0 +1,55 @@
+.\"
+.\" This file is part of RTRlib.
+.\"
+.\" This file is subject to the terms and conditions of the MIT license.
+.\" See the file LICENSE in the top level directory for more details.
+.\"
+.\" Website: http://rtrlib.realmv6.org/
+.\"
+.TH "rtrclient" "1"
+.SH NAME
+rtrclient \- rtr rpki client
+.SH SYNOPSIS
+.B rtrclient
+.B tcp
+[\fB\-k\fR]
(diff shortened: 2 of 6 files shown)


tools/rpki-rov.1
```diff
@@ -0,0 +1,28 @@
...
+.\" See the file LICENSE in the top level directory for more details.
+.\"
+.\" Website: http://rtrlib.realmv6.org/
+.\"
+.TH "rpki-rov" "1"
+.SH NAME
+rpki-rov - rpki route origin validator
+.SH SYNOPSIS
+.B rpki-rov
+[\fIhost\fR] [\fIport\fR]
```
waehlisch: Maybe we should make clear that "host" is the RPKI RTR cache server.

Btw, does the client connect by default to the well-known rpki-rtr TCP/323 port? If yes, we should make this clear and that the port is optional.

tools/rpki-rov.1
```diff
@@ -0,0 +1,28 @@
...
+.SH DESCRIPTION
+\fBrpki-rov\fR is a cli interface for rpki route origin validation.
```
waehlisch: s/cli interface for rpki route origin validation/command line interface that implemends RPKI route origin validation/

tools/rpki-rov.1
```diff
@@ -0,0 +1,28 @@
...
+It can be used interactively or via pipes.
+The format is always "\fIIP PREFIXLENGTH ASN\fR" followed by a newline.
```
waehlisch: The format of the input parameters is always ...

tools/rpki-rov.1
```diff
@@ -0,0 +1,28 @@
...
+.SH EXAMPLES
+Pipe into stdin
+.PP
+.nf
+.RS
+echo "93.175.146.0 24 12654" | rpki-rov rpki.example.com 8282
+.RE
+.fi
+.SH BUGS
+Does currently not support ssh based rpki server.
```
waehlisch: s/ssh based/ssh-based/

tools/rtrclient.1
```diff
@@ -0,0 +1,55 @@
...
+.B ssh
+[\fB\-k\fR]
+[\fB\-p\fR]
+.IR HOST
+.IR PORT
+.IR USERNAME
+.IR PRIVATE_KEY
+[\fIHOST_KEY\fR]
+.SH DESCRIPTION
+\fBrtrclient\fR allows to connect to an rtr server and print prefix and router key updates.
```
waehlisch: ... prints prefix, origin AS, and router key updates.

tools/rtrclient.1
```diff
@@ -0,0 +1,55 @@
...
+.SH OPTIONS
+\fB-k\fR
+.RS 4
+Print information about router key updates
+.RE
+\fB-p\fR
+.RS 4
+Print information about prefix updates
```
waehlisch: about prefix and origin AS.

tools/rtrclient.1
```diff
@@ -0,0 +1,55 @@
...
+.RE
+.SH EXAMPLES
+Print prefix updates from a tcp based server
```
waehlisch: (see above; valid ROA payloads consists of prefix AND origin AS.)

tools/rtrclient.1
```diff
@@ -0,0 +1,55 @@
...
+.PP
+.nf
+.RS
+rtrclient tcp -k rpki.example.com 8282
```
waehlisch: s/8282/323/

tools/rpki-rov.1
mroethke: How about adding "It connects to the rpki server described by \fIhost\fR and \fIport\fR and can be used …"
to the Description?
I believe that aligns well with other man pages.

waehlisch: Sounds good.

mroethke: done

PR #173, opened 2018-08-19
Do not include the RPATH in binaries.

lukas227:
Adding an RPATH is discouraged in Debian. See
https://wiki.debian.org/RpathIssue for reasoning.


waehlisch:
@lukas227 any opinion?

lukas227:
Merging #171 instead sounds good (sorry for not noticing that PR).

waehlisch:
Closed in favor of #171.

lukas227:
Sorry, I had the patch still locally applied when I (tried to) check whether #171 solves the rpath issue. It turns out that the patch is still needed to avoid generating an RPATH.

lukas227:
#171 actually does solve it. Sorry for the noise.

PR #175, opened 2018-08-20
Add description to CMake project() command

lukas227:
This description is used at least in the pkg-config file, where it is currently empty. (Feel free to change the description to whatever you like best.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,6 @@
-project(rtrlib C)
+project(rtrlib
+    DESCRIPTION "Lightweight C library that implements the RPKI/RTR protocol and prefix origin validation."
+    LANGUAGES C)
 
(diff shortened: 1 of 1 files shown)


mroethke:
LGTM

lukas227:
That seems a bit long; the longest decription in any of the pkg-config files on my workstation is 82 characters. How about a stripped down version: "Lightweight C library that implements the RPKI/RTR protocol.". For the contexts where this is displayed the fact that this is about RTRlib and that it is open source should be obvious. The second sentence seems a bit too detailed for this context to me.

I don't have hard feelings either way; if you want to go with your version I'll update the pull request.

waehlisch:
@lukas227 "Lightweight C library that implements the RPKI/RTR protocol and prefix origin validation." sounds good!

lukas227:
SGTM, updated pull request.

PR #176, opened 2018-08-21
Symbol visibility

### LINKED ISSUE #133: Public API and symbol visibility
Currently there is no clear definition of what we consider to be part of our public API and what not. Furthermore every symbol is public i.e. listed in the DSOs symbol table.
This makes "internal" changes somewhat problematic, because someone might use a function that they should not since we implicitly consider it not to be part of the public API.
This also makes it potentially difficult for users of rtrlib to rely on the stability of its API.

I therefore propose to clearly define a public API using a dedicated header file and symbol visibility i.e. hiding symbols that are not considered to be public.
The later can be accomplished by changing the default visibility to hidden using the compiler option `-fvisibility=hidden` and explicitly mark public functions with `__attribute__ ((visibility("default")))`.

mroethke:
This is a first attempt in defining a public api by setting the default to hidden and explicitly exporting the public api.
Please check if every relevant symbols is exported.

We furthermore need to change the header structure, because the including `rtrlib/rtrlib.h` currently pulls in most if not all headers, including symbols that would be considered private now. I will add that later to this PR.

Fix #133


mroethke:
I moved the export declarations from the header into the c files, because I think a library that uses rtrlib would otherwise reexport rtrlibs symbols.
And here is the list of the public symbols currently exported.
Note that `get_interval_mode` and `set_interval_mode` do not have a prefix, I will take care of that in a separate PR.

    get_interval_mode
    lrtr_ip_addr_to_str
    lrtr_ip_str_to_addr
    pfx_table_validate_r
    rtr_mgr_add_group
    rtr_mgr_conf_in_sync
    rtr_mgr_free
    rtr_mgr_get_spki
    rtr_mgr_init
    rtr_mgr_remove_group
    rtr_mgr_start
    rtr_mgr_status_to_str
    rtr_mgr_stop
    rtr_mgr_validate
    rtr_state_to_str
    set_interval_mode
    tr_close
    tr_free
    tr_open
    tr_ssh_init
    tr_tcp_init

waehlisch:
@fho, I would like to get some feedback from you on this issue. Thanks!

waehlisch:
ping @fho

mroethke:
Here is an updated symbol list, the only change being the interval_mode rename

    lrtr_ip_addr_to_str
    lrtr_ip_str_to_addr
    pfx_table_validate_r
    rtr_get_interval_mode
    rtr_mgr_add_group
    rtr_mgr_conf_in_sync
    rtr_mgr_free
    rtr_mgr_get_spki
    rtr_mgr_init
    rtr_mgr_remove_group
    rtr_mgr_start
    rtr_mgr_status_to_str
    rtr_mgr_stop
    rtr_mgr_validate
    rtr_set_interval_mode
    rtr_state_to_str
    tr_close
    tr_free
    tr_open
    tr_ssh_init
    tr_tcp_init

waehlisch:
@mroethke, can you resolve the conflicts? Did you test the new version with the current FRR integration?

rtrlib/lib/alloc_utils.h
```diff
@@ -12,6 +12,7 @@
 
 #include <stdlib.h>
 
+
```
smlng: unrelated

rtrlib/transport/tcp/tcp_transport.c
```diff
@@ -192,7 +193,7 @@ const char *tr_tcp_ident(void *socket)
     return sock->ident;
 }
 
-int tr_tcp_init(const struct tr_tcp_config *config, struct tr_socket *socket)
+RTRLIB_EXPORT int tr_tcp_init(const struct tr_tcp_config *config, struct tr_socket *socket)
```
smlng: this should (ideally) not be public

rtrlib/transport/transport.c
```diff
@@ -9,18 +9,19 @@
 
 #include "rtrlib/transport/transport.h"
 #include "rtrlib/lib/utils.h"
+#include "rtrlib/rtrlib_export.h"
 
-inline int tr_open(struct tr_socket *socket)
+RTRLIB_EXPORT inline int tr_open(struct tr_socket *socket)
```
smlng: these here shouldn't be public (IMHO) either

rtrlib/transport/transport.h
```diff
@@ -24,6 +24,7 @@
 #define RTR_TRANSPORT_H
 #include <time.h>
 
+
```
smlng: unrelated, too

rtrlib/lib/ip.c
```diff
@@ -46,14 +47,14 @@ bool lrtr_ip_addr_equal(const struct lrtr_ip_addr a, const struct lrtr_ip_addr b
     return lrtr_ipv4_addr_equal(&(a.u.addr4), &(b.u.addr4));
 }
 
-int lrtr_ip_addr_to_str(const struct lrtr_ip_addr *ip, char *str, const unsigned int len)
+RTRLIB_EXPORT int lrtr_ip_addr_to_str(const struct lrtr_ip_addr *ip, char *str, const unsigned int len)
```
smlng: I think this line should be broken into 2 or 3, to match 80 chars

rtrlib/pfx/trie/trie-pfx.c
```diff
@@ -336,7 +337,7 @@ inline void pfx_table_free_reason(struct pfx_record **reason, unsigned int *reas
         *reason_len = 0;
 }
 
-int pfx_table_validate_r(struct pfx_table *pfx_table, struct pfx_record **reason, unsigned int *reason_len, const uint32_t asn, const struct lrtr_ip_addr *prefix, const uint8_t prefix_len, enum pfxv_state *result)
+RTRLIB_EXPORT int pfx_table_validate_r(struct pfx_table *pfx_table, struct pfx_record **reason, unsigned int *reason_len, const uint32_t asn, const struct lrtr_ip_addr *prefix, const uint8_t prefix_len, enum pfxv_state *result)
```
smlng: please break this line too, to meet 80 chars

rtrlib/transport/tcp/tcp_transport.c
mroethke: Just like tr_ssh_init it must be, with the current api.

smlng: yeah, I know - hence _ideally_ 😉 

so leave as is

mroethke:
@waehlisch 
I rebased on master and frr does work now.

@smlng 
fixed

And here is an updated symbol list

    lrtr_ip_str_to_addr
    pfx_table_for_each_ipv4_record
    pfx_table_for_each_ipv6_record
    rtr_get_interval_mode
    rtr_mgr_add_group
    rtr_mgr_conf_in_sync
    rtr_mgr_free
    rtr_mgr_get_first_group
    rtr_mgr_get_spki
    rtr_mgr_init
    rtr_mgr_remove_group
    rtr_mgr_start
    rtr_mgr_status_to_str
    rtr_mgr_stop
    rtr_mgr_validate
    rtr_set_interval_mode
    rtr_state_to_str
    tr_ssh_init
    tr_tcp_init

waehlisch:
We still need one successful Travis run.

lukas227:
@mroethke the following symbols that are not in your list but exported:

lrtr_ip_addr_to_str
lrtr_set_alloc_functions
pfx_table_validate_r
rtr_mgr_for_each_ipv4_record
rtr_mgr_for_each_ipv6_record

mroethke:
@waehlisch 
you are right, however this is not quite finished yet anyway. I still need to finish restructuring the headers, because right now rtrlib.h contains symbols that we do not export anymore.

@lukas227 
Thank you for mentioning that, had an error in my regex.
So here is an updated list again, hopefully this is the last update.

    lrtr_ip_addr_to_str
    lrtr_ip_str_to_addr
    lrtr_set_alloc_functions
    pfx_table_for_each_ipv4_record
    pfx_table_for_each_ipv6_record
    pfx_table_validate_r
    rtr_get_interval_mode
    rtr_mgr_add_group
    rtr_mgr_conf_in_sync
    rtr_mgr_for_each_ipv4_record
    rtr_mgr_for_each_ipv6_record
    rtr_mgr_free
    rtr_mgr_get_first_group
    rtr_mgr_get_spki
    rtr_mgr_init
    rtr_mgr_remove_group
    rtr_mgr_start
    rtr_mgr_status_to_str
    rtr_mgr_stop
    rtr_mgr_validate
    rtr_set_interval_mode
    rtr_state_to_str
    tr_ssh_init
    tr_tcp_init

rtrlib/pfx/trie/trie-pfx.c
```diff
@@ -336,7 +337,14 @@ inline void pfx_table_free_reason(struct pfx_record **reason, unsigned int *reas
         *reason_len = 0;
 }
 
-int pfx_table_validate_r(struct pfx_table *pfx_table, struct pfx_record **reason, unsigned int *reason_len, const uint32_t asn, const struct lrtr_ip_addr *prefix, const uint8_t prefix_len, enum pfxv_state *result)
+RTRLIB_EXPORT int pfx_table_validate_r(
```
fho: Should we also export their variants? 
- `pfx_table_validate`
- `pfx_table_for_each_rec`

rtrlib/rtr/rtr.c
```diff
@@ -245,19 +246,19 @@ void rtr_stop(struct rtr_socket *rtr_socket)
     RTR_DBG1("Socket shut down");
 }
```
fho: I'm not sure if there is a good usecase to use rtr-sockets directly without the rtrmgr, but If we want to keep the option to use the rtr-sockets without the rtr-mgr, all of the the functions except the `static` ones needs to be exported.

rtrlib/pfx/trie/trie-pfx.c
mroethke: I don't think so.
The only reason I export `pfx_table_validate_r` is because there is nothing providing it's functionality on the mgr level and I don't see a use case for `pfx_table_for_each_rec`, as the only thing it has over the `for_each_ip*_record` functions is the ability to iterate a subtree, which does not make sense because nothing publicly visible returns a `trie_node`.

rtrlib/lib/ip.c
```diff
@@ -10,6 +10,7 @@
 #include <stdbool.h>
 #include <string.h>
 #include "rtrlib/lib/ip.h"
+#include "rtrlib/rtrlib_export.h"
```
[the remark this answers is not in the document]
mroethke: I agree.

rtrlib/rtr/rtr.c
mroethke: Using rtr_sockets without the manager would require initializing the tables yourself, which you cannot do anymore because neither `pfx_table_init` nor `spki_table_init` are exported.

fho: Yes, that would require to export more other symbols, probably most of the ones that are defined in the ".h" files.
Initially rtrmgr was only an additional wrapper that you could use if you want to use multiple rtr-sockets.
Depends on what the goal is, if rtrmgr should be the main interface to use the library.

rtrlib/pfx/trie/trie-pfx.c
fho: Maybe it's not much effort to provide a `pfx_table_validate_r` function on the rtrmgr level instead.

mroethke: sure, but I would not be comfortable with removing `pfx_table_validate_r` from the export table anyway. Because I believe the chances somebody actually uses this functions are rather high.

rtrlib/rtr/rtr.c
mroethke: I think providing rtr_mgr as the only interface should be goal at least long term.
It would make developing rtrlib much easier.

mroethke:
As a quick heads up, I have a working version with split headers that include only exported functions but it unfortunately breaks when rtrlib users include more than `rtrlib/rtrlibh.h`
which frr does. I have a solution for that but won't implement that today if anyone wants to take a look the current state can be found [here](https://github.com/mroethke/rtrlib/tree/symbol_visibility_wip).

rtrlib/pfx/trie/trie-pfx.c
waehlisch: @mroethke `pfx_table_validate_r` is used by `rpki-rov.c` for example.

mroethke:
This is now a log bigger than I though it would be but I believe this is mostly done now, apart from the travis build failure of course, which I cannot debug right now because it fails with an entirely different error on my machine.

Another review round is probably necessary though.

waehlisch:
Seems that we are almost close to merge. @fho can you give it another look?

PR #177, opened 2018-08-22
 rtr: prefix interval mode symbols 

mroethke:
The intervals mode symbols do not have a prefix at all right now. This
is problematic because it could easily lead to name collisions when used
in other projects.
Furthermore are the {set,get}_interval_mode functions in the wrong place
they operate on rtr_sockets and should therefore be in rtr.c not in
rtr_mgr.c

While this is a breaking API change I argue that it does not really
matter because the interval mode code has never officially been
released.

### DIFF
--- a/rtrlib/rtr/rtr.h
+++ b/rtrlib/rtr/rtr.h
@@ -51,26 +51,26 @@ enum rtr_rtvals {
     RTR_INVALID_PARAM = -2
 };
 
-enum interval_range {
-    BELOW_INTERVAL_RANGE = -1,
-    INSIDE_INTERVAL_RANGE = 0,
-    ABOVE_INTERVAL_RANGE = 1
+enum rtr_interval_range {
+    RTR_BELOW_INTERVAL_RANGE = -1,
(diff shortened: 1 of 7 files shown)


rtrlib/rtr/rtr.h
```diff
@@ -216,5 +216,20 @@ void rtr_stop(struct rtr_socket *rtr_socket);
  * @return !=NULL The rtr_socket_state as String.
  */
 const char *rtr_state_to_str(enum rtr_socket_state state);
+
+/**
+ * @brief Set the interval option to the desired one. It's either IGNORE_ANY,
+ * APPLY_ANY, DEFAULT_MIN_MAX or IGNORE_ON_FAILURE.
```
smlng: you may want to prefix this here, too?

smlng:
otherwise looks good!

colinbs:
Looks good to me!

waehlisch:
@salsh you also don't use it in BGPStream, right?

salsh:
@waehlisch you're right I don't use it

mroethke:
done

PR #178, opened 2018-08-26
Minor language correction

lukas227:
Replace "allows to" by "allows one to", since the tool does not set rules but enables people to do something.


waehlisch:
then, even better: "fBrtrclient\fR connects to an rtr server and prints prefix, origin AS and router key updates"

waehlisch:
print*s* ;)

tools/rtrclient.1
```diff
@@ -27,7 +27,7 @@ rtrclient \- rtr rpki client
 .IR PRIVATE_KEY
 [\fIHOST_KEY\fR]
 .SH DESCRIPTION
-\fBrtrclient\fR allows to connect to an rtr server and print prefix, origin AS and router key updates.
+\fBrtrclient\fR connects to an rtr server and print prefix, origin AS and router key updates.
```
smlng: is *rtr server* the correct term here? Maybe write:

```
connects to a RPKI cache server with RTR protocol and prints prefix, origin AS and router key updates. 
```

waehlisch: phew, now it gets complicated. Final version: "connects to an RPKI/RTR cache server and prints prefix, origin AS, and router key updates. "

lukas227: updated, including the oxford comma :)

PR #179, opened 2018-08-31
Fix Debian packaging.

sraustein:
This fixes several little things which were preventing the existing
Debian packaging from building cleanly with tools like gbp:

* There was a missing comma in one of the dependency lists in
  debian/control.

* debian/copyright was in a format that Lintian didn't like.

* The version numbering and source package format were wrong: the
  RTRlib project is the original source of this code, so this should
  have been a native package.  This patch changes the source package
  format from (implied) "1.0" to "3.0 (native)".

The change in source package format requires a corresponding change to
the package version numbering: native packages don't use the
"debian_revision" field of the package version number.

This patch adds a dummy 0.5.0.0 "release" so that the most recent
version in debian/changelog will be in the correct format and so that
the resulting package version numbers will sort correctly; feel free
to replace the "0.5.0.0" with the real 0.5.1 entry when that's ready.

A simpler alternative would have been just to retroactively whack all
the listed release numbers into the native format by removing the "-1"
suffix from every entry in debian/changelog, but that change could be
construed as rewriting history, so this patch takes the more
conservative approach.

See the "Version numbering" section of the Debian Policy Manual for
more than you ever wanted to know about Debian package versioning.

With these changes, the build works with gbp and passes Lintian.


debian/changelog
```diff
@@ -1,3 +1,11 @@
+librtr0 (0.5.0.0) stable; urgency=high
```
mroethke: would it not be better to increment epoch with the next release?

mroethke:
Why do you think that we should delete `debian/`?
I mean it does not hurt do keep it and it does make it easier to create you own rtrlib packages.

debian/changelog
sraustein: My understanding is that epoch bumps are the tool of last resort, to be used only when the upstream has screwed up the version numbering so badly that nothing else will work.  That does not seem to be the case here, so I didn't think it was necessary.

lukas227: We definitely should not increase epoch (that would be making it version number 1:0.5.0). I assume @mroethke  probably simply meant setting version to 0.5.1 with the next release.

mroethke: No, I meant epoch.

waehlisch:
all, do we converge here?

My personal opinion: I'm fine to remove `debian/` from the RTRlib directory to prevent confusion and inconsistent states.

@sraustein, @mroethke what do you prefer?

mroethke:
I still vote for keeping it because it is easier for people who want to build their own packages especially when building a specific commit and not a release. It is not a relevant maintenance burden either.

Removing it could also be a problem for frrs ci system, they currently build debian and ubuntu packages for their rpki enabled frr build.

And as a last point: I don't see how it would cause confusion as it is quite normal for upstream and downstream to differ in some ways.

waehlisch:
OK, we can also keep the pkg directory, it's not an either or.

As far as I see the only open question with this PR relates to the version number. From my point of view, the next release will be 0.6 as we did some larger changes.

PR #186, opened 2018-10-12
update to 0.6.1

mroethke:
This assumes #185 that will be merged first


waehlisch:
@mroethke before we merge: can you update README that C11 compiler is required.

Issue #184, opened 2018-10-12
rtrlib does not compile with pre c11 compilers since 0.6.0


mroethke:
We currently redefine two typedefs which is a feature that was introduces with c11.
The only relevant distribution that I'm aware is affected by this is centos6(and redhat of course).

The headers in question are `rtrlib/spki/hashtable/tommyds-1.8/tommytypes.h`, where the original definition is and `rtrlib/spki/hashtable/hashtable_forward.h` where the forward definition is. It is only needed for `rtr_mgr_config` from `rtr_mgr.h` which is a public header.

Possible solutions are (in no particular order)

1. Just include the header where the original types is. Of course that header is not enough because it has dependencies itself, which leads to a lot of bloat in our public header.
2. Wrap the `tommy_list` in `rtr_mgr_config` in something that can be forward declared. That is quite intrusive and leads to even more indirection when accessing groups. It would furthermore lead to an additional allocation because only pointer types can be forward declared for use in a struct.
3. Use a void pointer in `rtr_mgr_config` instead of the `tommy_list` type. This works but causes a lot of compiler warnings because tommy_list already is a pointer type. It thus looks like a double pointer. This could be solved with casts but casting pointer types like this is not exactly good practice. And of course this could not be type checked.
4. Officially adopt c11 and thereby drop support for centos7 and other long term support distributions that still use compilers without c11.
6. Replace `tommy_list` in `rtr_mgr_config` with something else (this is here just for completes sake).

I don't like any of these solutions but currently I would lean towards 2 or 4. Unless someone has a better idea.

waehlisch:
@fho opinions?

smlng:
Dropping CentOS7 (and then RH7, too) is not a good option, because there is no CentOS/RH8 (yet), so we would (as already stated) no longer support a major long-term stable distro.

fho:
> 5. Replace tommy_list in rtr_mgr_config with something else (this is here just for completes sake).

That might be a good option, it could be replaced with smaller less intrusive implementations:

The linked list can be replaced with an implementation similar to the one in the linux kernel:
- https://github.com/torvalds/linux/blob/master/include/linux/list.h
- https://github.com/tgraf/libnl/blob/master/include/netlink/list.h is a userspace variant of it without all the kernel macros

The hash table could be replaced with something like:
https://github.com/tgraf/libnl/blob/master/lib/hashtable.c

I don't know what the performance impacts will be

mroethke:
@smlng 
[...]

PR #188, opened 2018-10-23
rtr_mgr: wrap tommy_list in rtr_mgr_config in a struct

mroethke:
The purpose of this is to hide tommy symbols from public headers without
forward declaring tommy_list. Which is only possible with C11 as it is
typedefed.

#184


rtrlib/rtr_mgr.h
```diff
@@ -75,9 +74,11 @@ typedef void (*rtr_mgr_status_fp)(const struct rtr_mgr_group *,
 				  const struct rtr_socket *,
 				  void *);
 
+struct tommy_list_wrapper;
+
 //TODO Add refresh, expire, and retry intervals to config for easier access.
 struct rtr_mgr_config {
-	tommy_list groups;
+	struct tommy_list_wrapper *groups;
```
smlng: does this need to be a pointer?

mroethke: Yes, because tommy_list_wrapper due to the forward declaration the size of tommy_list_wrapper in rtr_mgr.h.

smlng: thanks, understood

mroethke:
ping @waehlisch

PR #191, opened 2018-10-25
debian packaging updates for 0.6.1

eqvinox:
After getting tired of waiting for someone of the Debian people to pick things up, I decided to go upload at https://mentors.debian.net/ myself :)

This is just a bunch of cleanups on the debian packaging.

NB: it would be very nice if you could put a release tag (v0.6.1.1) at commit 70874d6 so the version numbers match up (pretty please :D)

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -2,6 +2,7 @@ librtr0 (0.6.1.1) UNRELEASED; urgency=medium
 
   * add myself as maintainer as I'll be uploading this to mentors.
   * flip priorities from extra to optional (4.0.1 policy)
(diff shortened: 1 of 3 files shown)


waehlisch:
Thanks @eqvinox but actually we are almost done with this thanks to @lukas227, see https://salsa.debian.org/pkg-security-team/librtr

eqvinox:
> Thanks @eqvinox but actually we are almost done with this thanks to @lukas227, see https://salsa.debian.org/pkg-security-team/librtr

Even better, less work for me :D

PR #192, opened 2018-10-25
CMake: fix library installation path

eqvinox:
The GNUInstallDirs CMake module expands CMAKE_INSTALL_LIBDIR to a full
path, which breaks the path munging in CMakeLists.txt.  In particular, a
build on i386 Debian gets the install paths wrong like this:

`debian/tmp/build/librtr0-0.6.1.1/obj-i686-linux-gnu/lib/i386-linux-gnu/librtr.so.0.6.1`
instead of
`debian/tmp/usr/lib/i386-linux-gnu/librtr.so.0.6.1`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,6 +27,16 @@ if(NOT APPLE)
     find_library(RT_LIB rt)
 endif(NOT APPLE)
 
+# note: this must be before the include() directives below since
+# these directives will change CMAKE_INSTALL_LIBDIR to an absolute path
+if(IS_ABSOLUTE ${CMAKE_INSTALL_LIBDIR})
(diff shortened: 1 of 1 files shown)


waehlisch:
@mroethke looks similar to what @lukas227 was recently mentioned in a separate email?

mroethke:
@waehlisch 
yes, seems to be the same problem.

mroethke:
Reverting because it breaks debian {old,}stable, recent ubuntu and some rpm based builds.
Quite unfortunate that those that I tested are the ones that did not break …

https://ci1.netdef.org/browse/RPKI-RTRLIB-58

mroethke:
It seems this breaks debian {old,}stable, recent ubuntu and some rpm based builds.
The ones I did test are among the few tat did not break …

https://ci1.netdef.org/browse/RPKI-RTRLIB-58

Issue #190, opened 2018-10-25
Missing exposure of structs and functions


waehlisch:
The RPKI integratoin into BGPstream (see also https://github.com/CAIDA/libbgpstream/pull/58) requires exposure of:

*  struct pfx_table
*  pfx_table_init()
*  pfx_table_add()
*  pfx_table_free()
*  pfx_table_src_remove()

cc @salsh

salsh:
Thanks for opening the issue.

salsh:
The ROAFetchlib validates historical BGP dumps with historical ROA dumps. For the validation process multiple prefix tables are initialized, the entries of the ROA dumps are added as records, the actual validation process of the BGP entries with the prefix tables takes place and and finally, the prefix tables are removed and freed.

waehlisch:
> I don't think the pfx_table should be used directly and independently of rtr_mgr.

@mroethke indeed it depends on the use case ;). RTRlib basically provides two functions: (1) data structure and ROV algorithm, (2) RTR protocol. There may be good reasons why people use RTRlib just because of (1), see comment by @salsh. I think we should expose the mentioned functions.

mroethke:
alright.

PR #194, opened 2018-10-26
pfx: expose enough of pfxt to make it independently usable

mroethke:
This exposes enough of pfxt to make it usable without rtr_mgr.

#190 
@salsh


waehlisch:
@salsh can you please test.

salsh:
It's working. Thank you.

Issue #147, opened 2017-12-02
cmake CMAKE_INSTALL_LIBDIR sometimes absolute path


mwinter-osr:
Something I ran into, trying to get the RPM package working on CentOS 6. I have close to zero cmake experience, so need advise from someone:

If I build on CentOS 7 (and other newer systems):, the rtrlib.pc ends up like this:
```prefix=/usr
includedir=${prefix}/include
libdir=${prefix}/lib64

Name: rtrlib
Description: 
Version: 0.5.0
Requires: libssh >= 0.5.0
Libs: -L${libdir} -lrtr
Cflags: -I${includedir}
```

However on a CentOS 6 system, I get a wrong path for `libdir`:
```prefix=/usr
includedir=${prefix}/include
libdir=${prefix}//usr/lib64

Name: rtrlib
Description: 
Version: 0.5.0
Requires: libssh >= 0.5.0
Libs: -L${libdir} -lrtr
Cflags: -I${includedir}
```

This ends up with `libdir=/usr//usr/lib64` instead of `libdir=/usr/lib64`

The issue is based on `CMakeLists.txt`:
```
set (PKG_CONFIG_LIBDIR     "\${prefix}/${CMAKE_INSTALL_LIBDIR}")
```

However, `CMAKE_INSTALL_LIBDIR` is **NOT** always a relative path. Older systems seem to provide an absolute path.
See discussion at 
https://bugzilla.redhat.com/show_bug.cgi?id=1301720
and https://cmake.org/pipermail/cmake/2012-February/049250.html

Any thoughts on how to resolve this?

waehlisch:
@fho any idea?

fho:
@waehlisch no, sorry :-)

waehlisch:
Hmm, I also don't have a good idea. It's about a workaround of an old bug. @smlng?

@mwinter-osr what do you prefer?

mwinter-osr:
I've tried some workaround, but so far with no success. I'll try a bit more over the weekend.
But I'm a cmake novice (never used cmake before). I was counting on someone here understanding cmake better than I do.

mroethke:
It seems the old cmake version is the problem.
I have found two possible solutions

1. Ship the cmake module GNUInstallDirs and include it if necessary. 
2. Just check if CMAKE_INSTALL_LIBDIR is absolute and handle it accordingly.

I did not check if the first one acualty works, the second one is implemented in #150

waehlisch:
[...]

PR #196, opened 2018-11-03
Revert "CMake: fix library installation path (#192)"

mroethke:
This reverts commit 27395a52f3055e4691d08fb8c37bed6006c0db65.

It breaks much more than it fixes

@eqvinox

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,16 +27,6 @@ if(NOT APPLE)
     find_library(RT_LIB rt)
 endif(NOT APPLE)
 
-# note: this must be before the include() directives below since
-# these directives will change CMAKE_INSTALL_LIBDIR to an absolute path
-if(IS_ABSOLUTE ${CMAKE_INSTALL_LIBDIR})
-    set (PKG_CONFIG_LIBDIR ${CMAKE_INSTALL_LIBDIR})
-    set (RTRLIB_INSTALL_LIBDIR ${CMAKE_INSTALL_LIBDIR})
-else()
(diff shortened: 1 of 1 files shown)


eqvinox:
```
02-Nov-2018 22:38:34	-- Installing: /home/ci/cibuild.58/rtrlib-source/debian/tmp/usr/pkgconfig/rtrlib.pc
```

Well... `CMAKE_INSTALL_LIBDIR` isn't set before the `include` directives are evaluated below.

The correct fix is probably to remove the entire `if(IS_ABSOLUTE ${CMAKE_INSTALL_LIBDIR})` block.

waehlisch:
@mroethke can you check the new proposal?

mroethke:
That code was added because of #147.
Since centos 6 is still relevant I would not consider completely removing it as a viable option

eqvinox:
So — I don't really know cmake well, and after looking at #147 I'm even more confused now.  In my tests, `CMAKE_INSTALL_LIBDIR` was _always_ an absolute path after the `include(UseMultiArch)` and `include(GNUInstallDirs)` statements were evaluated.  #147 seems to indicate the opposite, so either I mis-tested/misinterpreted something, or something changed in the meantime since #147 was filed.

That said, the situation is now a choice between making CentOS 6 work and making current Debian work (on i386).  I'd try to go for the "best" fix on current systems, and then on top of that fix CentOS 6 in some way that doesn't break other things.  (Passing some different value of `CMAKE_INSTALL_LIBDIR` on the cmake command line on CentOS 6 maybe?)

waehlisch:
@mwinter-osr, @eqvinox can you check that it works for you?

waehlisch:
Closed in favor of #197.

PR #198, opened 2018-11-16
tests: add connect timeout to test_dynamic_groups

mroethke:
Without a timeout the test will hang indefinitely if for whatever reason
the connection to the cache server can not be established.

### DIFF
--- a/tests/test_dynamic_groups.c
+++ b/tests/test_dynamic_groups.c
@@ -7,6 +7,17 @@
 #include "rtrlib/rtr_mgr_private.h"
 #include "rtrlib/spki/hashtable/tommyds-1.8/tommylist.h"
 
+const int connection_timeout = 20;
+enum rtr_mgr_status connection_status = -1;
+
+static void connection_status_callback(const struct rtr_mgr_group *group,
+				       enum rtr_mgr_status status,
+				       const struct rtr_socket *socket,
+				       void *data)
+{
+	connection_status = status;
+}
+
 int main(void)
 	{
 	//create a TCP transport socket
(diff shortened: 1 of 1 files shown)


waehlisch:
@smlng does this PR looks fine to you?

tests/test_dynamic_groups.c
```diff
@@ -7,6 +7,18 @@
...
 
+const int connection_timeout = 20;
+enum rtr_mgr_status connection_status = -1;
+
+static void connection_status_callback(const struct rtr_mgr_group *group,
+				       enum rtr_mgr_status status,
+				       const struct rtr_socket *socket,
+				       void *data)
+{
+	if (status == RTR_MGR_ERROR)
```
smlng: is this necessary? Why not just update the status all the time, the check for error is done in the loop below anyway.

tests/test_dynamic_groups.c
```diff
@@ -46,14 +58,23 @@ int main(void)
...
 	//start the connection manager
 	rtr_mgr_start(conf);
 
-	//wait till at least one group is fully synchronized with the server
-	while (!rtr_mgr_conf_in_sync(conf))
+	int sleep_counter = 0;
+	// wait 20 sec till at least one group is fully synchronized with the server
+	// otherwise EXIT_FAILURE.
+	while (!rtr_mgr_conf_in_sync(conf)) {
+		if (connection_status == RTR_MGR_ERROR)
```
smlng: maybe nit picky, but I would write this as:

```C
while (!rtr_mgr_conf_in_sync(conf)) {
    sleep_counter++;
    if ((connection_status == RTR_MGR_ERROR) ||
        (sleep_counter > connection_timeout))
            return EXIT_FAILURE;
    sleep(1);
}
```

as the sleep comes last this should only sleep 20 times at most, or exit before, thus avoiding the `>=` and also merging both error conditions.

mroethke:
CI:rerun

waehlisch:
As soon as this PR is fixed and merged, we will publish a new release.

mroethke:
changed as requested by @smlng and rebased to resolve CI errors

waehlisch:
Ping @smlng.

PR #199, opened 2018-11-23
README: remove c11 requirement

mroethke:
C11 was only required because of typedef forward declarations, but they
are not used anymore (see PR #188).


mroethke:
CI:rerun

PR #213, opened 2019-02-05
rtrlib/transport: fix ssh transport

### LINKED ISSUE #211: SSH support broken
SSH support is broken since fa601d1a175700d3f4c1c9b0782f07e0822f256a.
Unless a bind address is specified private key and host key will not be saved in the configuration.

mroethke:
Currently the ssh transport will never connect successfully because
unless a bindaddr is specified.
This also fixes hostkey verification.

fix #211


mroethke:
CI:rerun

PR #214, opened 2019-02-06
rtrlib/rtr: fix race condition in rtr_stop

mroethke:
rtr_stop currently has a race condition. It uses a unsynchronized shared
variable. Which value this variable has is therefore undefined when the
thread that rtr_stop tries to kill reads it. The only direct solution is
a explicit memory barrier, but C does not support them until C11, even
there it's optional.

The new approach uses pthread_cancel and carefully selected code regions
where it is safe to exit without freeing resources. The shared variable
is still used but we do not rely on it for timely thread exit anymore.

### DIFF
--- a/rtrlib/rtr/packets.c
+++ b/rtrlib/rtr/packets.c
@@ -168,6 +168,15 @@ struct pdu_end_of_data_v1 {
     uint32_t expire_interval;
 };
 
+struct recv_loop_cleanup_args {
+    struct pdu_ipv4 *ipv4_pdus;
+    struct pdu_ipv6 *ipv6_pdus;
+    struct pdu_router_key *router_key_pdus;
+};
+
+
+static void recv_loop_cleanup(void *p);
+
 static int rtr_send_error_pdu_from_network(const struct rtr_socket *rtr_socket,
                                            const void *erroneous_pdu,
                                            const uint32_t erroneous_pdu_len,
@@ -991,6 +1000,15 @@ static int rtr_update_spki_table(struct rtr_socket* rtr_socket, struct spki_tabl
     return RTR_SUCCESS;
 }
 
+
+void recv_loop_cleanup(void *p) {
+    struct recv_loop_cleanup_args *args = p;
+    lrtr_free(args->ipv4_pdus);
+    lrtr_free(args->ipv6_pdus);
+    lrtr_free(args->router_key_pdus);
+}
+
+
 int rtr_sync_receive_and_store_pdus(struct rtr_socket *rtr_socket){
     char pdu[RTR_MAX_PDU_LEN];
     enum pdu_type type;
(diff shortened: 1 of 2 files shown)


nbars:
Hey, some remarks from my side:

1. If we are not in the `cancelable section`, we might block up to 60 seconds while receiving `PDUs`. Furthermore, this blocking period might further increase, if we are talking to a server that responses very slowly. I think it is not acceptable for a user of our library to wait such long for a shutdown.
2. As `pthread_cancel` causes the target thread to terminate at `cancelation points,` we have no way to clean up the environment; thus this might cause resources to leak. Although they might be no allocations as of now in the `cancelable section`, this might change in further commits (and it is not apparent to a developer that this can cause leaks).

mroethke:
1. That already is a problem with the current solution. Because the signal might arrive while not receiving and `rtr_socket->state` is never checked.
This patch at least improves the situation because currently it may block for up to sync_interval in `rtr_wait_and_sync`, which may be hours or even a full day.

2. Not True. Yes it may terminate at any cancellation point within a cancellable section, but you can use `pthread_cleanup_{push,pop}` for cleanup. And about this not being "apparent", that's what documentation is for.

How about this:
- Add explicit cancellation point (`pthread_testcancel`) to the end of the fsm loop
- use `pthread_cancel` followed by `pthread_kill`

This way we will return from blocking calls and do not need to worry about synchronisation because with `pthread_cancel` the runtime does this for us. Furthermore we don't have to worry about resources because at the end of the fsm loop no resources are held that are not accessible via the rtr_socket struct.

mroethke:
ping @nbars

nbars:
Doesn't this approach void your first point that we are improving this situation?
If the signal arrives just before we are sleeping for hours, this will not work anymore.

Maybe we could call `shutdown` on the socket before sending the signal?

I prefer to use `select/poll` on the socket `fd` and a `fd` that is used by the `main thread` to notify the other thread that it should stop. However, without looking at the code, I believe that this might be a more significant change (or not even possible with the SSH lib).

mroethke:
Yes, unless we keep the `cancelable section` around `rtr_wait_for_sync`.

I don't think using `shutdown` would work. It seems the socket does not block on recv anymore, but does not give any indication that it cant receive data anymore.

select/pool would work with libssh, we just have to open the socket our self and set it with `ssh_options_set`.
But *with* looking at the code: That would not be trivial at all. The FSM would need significant changes. I think we should to this anyway, but that is to intrusive for a "quick fix".

Making both `rtr_receive_pdu` calls a `cancelable section`  and ensuring cleanup using `pthread_cleanup_{push,pop}` might be the best solution for now.

mroethke:
@nbars can you take another look?

rtrlib/rtr/rtr.c
```diff
@@ -240,10 +244,17 @@ void rtr_stop(struct rtr_socket *rtr_socket)
     RTR_DBG1("rtr_stop()");
     rtr_change_socket_state(rtr_socket, RTR_SHUTDOWN);
     if(rtr_socket->thread_id != 0) {
-        RTR_DBG1("pthread_kill()");
-        pthread_kill(rtr_socket->thread_id, SIGUSR1);
```
nbars: The sig handler is now dead code, right?

rtrlib/rtr/packets.c
```diff
@@ -991,6 +991,20 @@ static int rtr_update_spki_table(struct rtr_socket* rtr_socket, struct spki_tabl
...
 
+struct cleanup_args {
+    struct pdu_ipv4 *ipv4_pdus;
+    struct pdu_ipv6 *ipv6_pdus;
+    struct pdu_router_key *router_key_pdus;
+};
+
+void cleanup(void *p) {
+    struct cleanup_args *args = p;
+    free(args->ipv4_pdus);
```
nbars: `lrtr_free`

rtrlib/rtr/rtr.c
mroethke: yes, removed.

rtrlib/rtr/packets.c
mroethke: fixed.

mroethke:
I added warnings to the functions containing cancelable sections, not sure if that is enough though. Because functions called in those sections are affected to of course. For some functions it is rather obvious that they will never allocate resources …

mroethke:
ci:rerun

mroethke:
CI:rerun

mroethke:
ping @nbars

PR #225, opened 2019-03-13
Travis ci updates

mroethke:
### Contribution description
- Update checkpatch
- fix issues found by new checkpatch
- Check the diff in addition to the whitelisted files with checkpatch
This will not (yet) fail the build because it could cause problems with changes files not conforming to our standards.
- Ensure all build steps are executed even if some fail


mroethke:
I think with parentheses is better, because it makes clear it references a function without needing additional words.

PR #229, opened 2019-04-01
Improve ssh build time options and output

mroethke:
### Contribution description
Currently the build type has to be changed to disable ssh support, that is suboptimal.
Furthermore the output was somewhat confusing when ssh support was disabled.

This also re indents FindLibSSH.cmake, as it was quite hard to read.

### Testing procedure
- Run cmake $BUILD_OPTIONS
    - Works regardless of libssh presence but warns without it
- Run cmake $BUILD_OPTIONS -DRTRLIB_TRANSPORT_SSH=Yes
    - Works only when a supported libssh was found
- Run cmake $BUILD_OPTIONS -DRTRLIB_TRANSPORT_SSH=No
    - Works regardless of libssh presence and does not warn without it


Arvedui:
done

mroethke:
done

mroethke:
CI:rerun

Issue #219, opened 2019-02-20
Extended rtrlclient to create multiple outputs


waehlisch:
#### Background
Multiple implementations of RPKI origin validation (e.g., OpenBGPD) don't use RPKI RTR to maintain valid ROA data but import data from external files. Therefore, some RPKI validators provide a corresponding output which is then fed to the BGP router on a regular basis.

Such validator-centric approach is not good operational practice as it binds ROA output to a single Validator implementation and running instance. It would be better to benefit from the RPKI architecture, which allows a BGP router to receive data from multiple relying parties, e.g., to improve reliability.

#### Proposal (1)
The proposal is to extend the rtrclient such that it creates multiple output formats (see below) of the received valid ROA payload. It seems also helpful to introduce an option that terminates the rtrclient after receiving all data to make it easier to use in cron settings.

#### Proposal (2)
Preferably, we allow for easy extension of the output format, i.e., the user can configure how the output should look like. Typically, the required output consists of a fix preamble, dynamic content that includes the ROA, and a postamble.

For example, the OpenBGPD config:
```
roa-set { 
    203.119.88.0/23 maxlen 24 source-as 187
    2001:dd8:7::/48 maxlen 48 source-as 187        
    2401:680::/32 maxlen 32 source-as 715
}
deny quick from ebgp ovs invalid
```

A configuration file for this might look like
```
peamble:
  roa-set { 
content:
  $prefix/$length maxlen $maxlen source-as $origin
postamble:
  }
deny quick from ebgp ovs invalid
```

For a CSV file
```
  preamble:
    #this is a ROA CSV file
  content:
   $prefix,$length,$maxlen,$origin
  postamble:
```

### Useful links
https://rpki.readthedocs.io/en/latest/routinator/running.html#printing-a-list-of-valid-route-origins

mroethke:
This seems out of scope for rtrclient, because what you want is to dump the state of the pfx table after the first sync. rtrclient just dumps the diff it applies to the table, that would not make much sense for the proposed use case especially when used with multiple caches.
This should be its own tool.

I would furthermore like to avoid implementing our own templating engine, there are more than enough out there already. [mustache](https://mustache.github.io/) [(c implementation)](https://gitlab.com/jobol/mustach) might be a candidate.

A script language might also be better suited for this than C.

waehlisch:
[...]

PR #233, opened 2019-05-07
tools/rtrclient: support multiple caches

mroethke:
### Contribution description

Add support for multiple caches in rtrclient.

### Testing procedure
Try to specify multiple cache server like this:
`rtrclient tcp rpki-validator.realmv6.org 8282 tcp rpki-validator.realmv6.org 8283`
Debug output should indicate two connections.

### Issues/PRs references
Preparation for #219


mroethke:
Travis fails because of checkpatch. I'm not sure what to do about that. 
Breaking the lines causes more warnings because kernel coding style says one should not break user visible strings.

tools/rtrclient.c
```diff
@@ -7,35 +7,188 @@
...
+
+/**
+ * @brief Return value checking realloc wrapper. Exits program on error.
+ */
+static inline void *checked_realloc(void *ptr, size_t size)
+{
+	return check_alloc_ret(realloc(ptr, size));
+}
+
+static struct socket_config *extend_socket_config() {
```
smlng: brace should be on new line

tools/rtrclient.c
```diff
@@ -7,35 +7,188 @@
...
+	for (size_t i = 0; i < strlen(str); ++i) {
+		if (!isdigit(str[i]))
+			return false;
+	}
+	return true;
+}
+
+static bool is_valid_port_number(const char *str)
+{
+	if (!is_numeric(str))
```
smlng: this check can be omitted by using `strtol` (or even better `strtoul`) instead of `atoi` below, on error it returns `0` which is not a valid port number anyway.

mroethke: unfortunately strtol ignores trailing characters, my check does not.

smlng:
please rebase

tools/rtrclient.c
smlng: okay

smlng:
and squash

smlng:
can we somehow disable the error by checkpatch when initialising globals with default values?

PR #238, opened 2019-05-21
scripts/travis: fix syntax error

mroethke:
### Contribution description
The patch to master should be displayed for PR runs on travis if the style checker found issues. This does not work right now because of a syntax error.


mroethke:
netdef build finished successfully 30 minutes ago, it just did not generate a report for some reason.

PR #241, opened 2019-06-06
Implement ROA export

### LINKED ISSUE #219: Extended rtrlclient to create multiple outputs

mroethke:
### Contribution description

Implement export of ROA information through rtrclient in arbitrary formats.

### Issues/PRs references

Fix #219


smlng:
without diving into this too deep, but from first thought: can't we integrate the `mustach` library differently than copying the whole project (including files like `.gitignore` and `.gitlab-ci.yml`). For instance by using CMAKE external projects, see [here](https://cmake.org/cmake/help/v3.8/module/ExternalProject.html)?

waehlisch:
Don't make it complicate in terms of tooling.

mroethke:
I removed of the mustachs files, keeping only what we use + license, readme and authors files.

The remaining travis issue is fixed by #242, which should probably be merged first.

tools/rtrclient.c
```diff
@@ -98,6 +116,60 @@ static struct socket_config *extend_socket_config()
...
+
+	if (is_readable_file(name)) {
+		FILE *template_file = fopen(name, "r");
+
+		if (fseek(template_file, 0, SEEK_END) == -1)
+			goto read_error;
+
+		long file_size = ftell(template_file);
+		if (file_size == -1)
+			goto read_error;
```
smlng: close file?

tools/rtrclient.c
```diff
@@ -98,6 +116,60 @@ static struct socket_config *extend_socket_config()
...
+
+		rewind(template_file);
+
+		char *template = checked_malloc(file_size);
+
+		size_t bytes_read = fread(template, 1, file_size, template_file);
+
+		if (bytes_read != (unsigned long) file_size)
+			goto read_error;
```
smlng: close file?

tools/rtrclient.c
```diff
@@ -140,19 +212,144 @@ static bool is_resolveable_host(const char *str)
 	return true;
 }
 
-#ifdef RTRLIB_HAVE_LIBSSH
+static bool is_file(const char *str)
+{
+	struct stat stat_buf;
+
+	if (stat(str, &stat_buf) == -1 )
```
smlng: remove space after `-1`

smlng: travis (check patch in there) found that too and one or two more

tools/rtrclient.c
```diff
@@ -169,6 +367,10 @@ static void print_usage(char **argv)
 	printf("-k  Print information about SPKI updates.\n");
 	printf("-p  Print information about PFX updates.\n\n");
 
+	printf("-e export pfx table and exit\n");
```
smlng: these are not reflected/shown in `usage` nor `examples` in the help message.

Also: please change port from `8282` to `8283` in examples to utilise the RIPE Validator v3

tools/rtrclient.c
mroethke: done

tools/rtrclient.c
mroethke: same

tools/rtrclient.c
mroethke: already done

tools/rtrclient.c
mroethke: I added examples for the export functionality and documented the `-o` option.

third-party/CMakeLists.txt
```diff
@@ -1 +1,4 @@
 set(tommyds "${CMAKE_CURRENT_SOURCE_DIR}/tommyds/tommy.c" PARENT_SCOPE)
+
+set(mustach "${CMAKE_CURRENT_SOURCE_DIR}/mustach/mustach.c" PARENT_SCOPE)
+set_source_files_properties("${CMAKE_CURRENT_SOURCE_DIR}/mustach/mustach.c" PROPERTIES COMPILE_FLAGS "-DNO_EXTENSION_FOR_MUSTACH")
```
smlng: why is this set here and again in `tools/CMakeLists.txt`, are both required? Btw. why not for tommyds then?

tools/rtrclient.c
```diff
@@ -167,15 +368,21 @@ static void print_usage(char **argv)
 	printf("-b  bindaddr Hostnamne or IP address to connect from\n\n");
 
 	printf("-k  Print information about SPKI updates.\n");
-	printf("-p  Print information about PFX updates.\n\n");
+	printf("-p  Print information about PFX updates.\n");
+	printf("-s  Print information about connection status updates.\n\n");
```
smlng: this is missing in usage above

third-party/CMakeLists.txt
mroethke: I probably just to forgot to remove that one …

mroethke:
not because no format is defined, as there is a default format, but because `-e` is not defined. Same for `-t`.

mroethke:
That is exactly what it did when I entered the command just now

smlng:
now it prints usage instead of running, however it does not state why its failing - can we add a reasonable error message (`invalid template`) before printing the help.

```
rtrclient -e -t foo -o roa.foo tcp rpki-validator.realmv6.org 8283
Usage:
 rtrclient tcp [options] <host> <port>
 rtrclient ssh [options] <host> <port> <username> <private_key> [<host_key>]

Options:
-k  Print information about SPKI updates.
-p  Print information about PFX updates.

Examples:
 rtrclient tcp rpki-validator.realmv6.org 8282
 rtrclient tcp -k -p rpki-validator.realmv6.org 8282
 rtrclient ssh rpki-validator.realmv6.org 22 rtr-ssh ~/.ssh/id_rsa ~/.ssh/known_hosts
 rtrclient ssh -k -p rpki-validator.realmv6.org 22 rtr-ssh ~/.ssh/id_rsa ~/.ssh/known_hosts
```

smlng:
please squash fixups

smlng:
tested ACK, please squash

mroethke:
done

PR #245, opened 2019-07-08
tools: generate template.h in build folder

smlng:
### Contribution description

The template.h is generated when building the rtr tools.
The resulting file is now create in the build and not the
source directory. This avoids that template.h shows up in
`git status` and also that someone might commit the file
in to the repo by accident.

### Testing procedure

Build rtrlib on master:

```
mkdir build
cd build
cmake ../
```

Then run `git status` which will show a new file `tools/template.h` that was
created.

Now do the same with this PR and you'll that this file is not created in `tools/` anymore, 
but in `build/tools/`.

### DIFF
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -32,5 +32,6 @@ endforeach()
 
 string(STRIP ${TEMPLATES} TEMPLATES)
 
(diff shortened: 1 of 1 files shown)


smlng:
see https://cmake.org/cmake/help/latest/command/configure_file.html

PR #248, opened 2019-09-18
*: fix warnings of compilation with gcc 5.4

pguibert6WIND:
some warnings are removed by adding some tiny modifications in the code.

Signed-off-by: Philippe Guibert <philippe.guibert@6wind.com>

### DIFF
--- a/tests/test_live_validation.c
+++ b/tests/test_live_validation.c
@@ -14,15 +14,15 @@
 #include <string.h>
 #include <unistd.h>
 
-#define RPKI_CACHE_HOST "rpki-validator.realmv6.org"
-#define RPKI_CACHE_POST "8283"
-
 struct test_validity_query {
-	char *pfx;
+	const char *pfx;
 	int len;
 	int asn;
-	int val;
+	unsigned int val;
 };
+
+static char *RPKI_CACHE_HOST = "rpki-validator.realmv6.org";
+static char *RPKI_CACHE_POST = "8283";
 
 /*
  * Verification is based on ROAs for RIPE RIS Routing Beacons, see:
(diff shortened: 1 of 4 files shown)


third-party/mustach/mustach.c
```diff
@@ -60,7 +60,8 @@ static int memfile_close(FILE *file, char **buffer, size_t *size)
 	return rc;
 }
 #else
-static FILE *memfile_open(char **buffer, size_t *size)
+static FILE *memfile_open(char **buffer __attribute__((unused)),
```
mroethke: I would prefer not to change third-party libs without going thorough upstream

tools/rpki-rov.c
```diff
@@ -19,8 +19,10 @@
 const int connection_timeout = 20;
 enum rtr_mgr_status connection_status = -1;
 
-static void connection_status_callback(const struct rtr_mgr_group *group, enum rtr_mgr_status status,
-				       const struct rtr_socket *socket, void *data)
+static void connection_status_callback(const struct rtr_mgr_group *group __attribute__((unused)),
+                                       enum rtr_mgr_status status,
```
mroethke: checkpatch is still not happy with the indention here, running `clang-format -i tools/rpki-rov.c` will fix this.

pguibert6WIND: sorry, could you be more specific about the issue I should do:
calling clang-format resulted with some errors
https://gist.github.com/pguibert6WIND/c57c562250f74630e6d79741dc985f61

indeed, I use clang-format 3.8, and I read that I had to move to clang-format 4.0. this is what I said. otherwise I would have had that:
```
YAML:32:21: error: unknown key 'IncludeIsMainRegex'
IncludeIsMainRegex: '(private)?$'
                    ^~~~~~~~~~~~~
Error reading /root/ENV/rtrlib/.clang-format: Invalid argument
```

mroethke: lines 23 and 24 are indented with spaces only, where kernel coding style says you should tabs where possible and spaces when necessary.
Take a look at line 25 for instance.

pguibert6WIND: done. thanks

tests/test_live_validation.c
```diff
@@ -57,7 +59,7 @@ int main(void)
 {
 	/* create a TCP transport socket */
 	struct tr_socket tr_tcp;
-	struct tr_tcp_config tcp_config = {RPKI_CACHE_HOST, RPKI_CACHE_POST, NULL};
+	struct tr_tcp_config tcp_config = { (char *)RPKI_CACHE_HOST, (char *)RPKI_CACHE_POST, NULL};
```
mroethke: I'm not sure casting away const is the best fix here.
Copying the string would work.
@smlng your opinion?

mroethke: After RL discussion with @smlng  we think it would be best to declare `RPKI_CACHE_HOST` and `RPKI_CACHE_POST` as static variables (and maybe fix the typo while touching them …)

pguibert6WIND:
please review. I made changes

smlng:
looks good to me now

Issue #212, opened 2019-02-05
SSH transport does not honor sync timeout


mroethke:
rtrlib implements the sync timeout by blocking on receive for the remainder of the timeout. The SSH transport does not implement this at all. It returns immediately causing rtrlib to send a serial query way to often (about once per second).

PR #256, opened 2020-06-16
SSH fixes

### LINKED ISSUE #250: Extend SSH transport to allow for password authentication
### Description of current state
In case of SSH transport, only key-based authentication is allowed.

### Improvement
Allow also for password-based authentication.

### Useful links
Can be tested using https://github.com/cloudflare/gortr.

mroethke:
### Contribution description
Several fixes for the ssh transport
- the log prefix did sometimes indicate the wrong function
- add support for password authentication
- properly check malloc return values in ssh init
- fix a deprecation warning from libssh
- fix constant spinning on send
- check return values of libssh functions better and translate them explicitly to our internal ones

### Issues/PRs references
fixes #212 
fixes #250

### DIFF
--- a/rtrlib/transport/ssh/ssh_transport.c
+++ b/rtrlib/transport/ssh/ssh_transport.c
@@ -299,6 +299,9 @@ RTRLIB_EXPORT int tr_ssh_init(const struct tr_ssh_config *config, struct tr_sock
 	ssh_socket->config.port = config->port;
 	ssh_socket->config.username = lrtr_strdup(config->username);
 
+	if ((config->password && config->client_privkey_path) || (!config->password && !config->client_privkey_path))
+		return TR_ERROR;
+
 	if (config->bindaddr)
 		ssh_socket->config.bindaddr = lrtr_strdup(config->bindaddr);
 	else
--- a/rtrlib/transport/ssh/ssh_transport.c
+++ b/rtrlib/transport/ssh/ssh_transport.c
@@ -137,14 +137,29 @@ int tr_ssh_open(void *socket)
 		goto error;
 	}
 
+	if (config->client_privkey_path) {
+		SSH_DBG("%s: Trying publickey authentication", ssh_socket, __func__);
+		ssh_options_set(ssh_socket->session, SSH_OPTIONS_IDENTITY, config->client_privkey_path);
+
+		int rtval;
+
 #if LIBSSH_VERSION_MAJOR > 0 || LIBSSH_VERSION_MINOR > 5
-	const int rtval = ssh_userauth_publickey_auto(ssh_socket->session, NULL, NULL);
+		rtval = ssh_userauth_publickey_auto(ssh_socket->session, NULL, NULL);
 #else /* else use libSSH version 0.5.0 */
-	const int rtval = ssh_userauth_autopubkey(ssh_socket->session, NULL);
+		rtval = ssh_userauth_autopubkey(ssh_socket->session, NULL);
 #endif
-	if (rtval != SSH_AUTH_SUCCESS) {
-		SSH_DBG("%s: Authentication failed", ssh_socket, __func__);
-		goto error;
+		if (rtval != SSH_AUTH_SUCCESS) {
+			SSH_DBG("%s: Publickey authentication failed", ssh_socket, __func__);
+			goto error;
+		}
+
+	} else {
+		SSH_DBG("%s: Trying password authentication", ssh_socket, __func__);
+
+		if (ssh_userauth_password(ssh_socket->session, NULL, config->password) != SSH_AUTH_SUCCESS) {
+			SSH_DBG("%s: Password authentication failed", ssh_socket, __func__);
+			goto error;
+		}
 	}
 
 	ssh_socket->channel = ssh_channel_new(ssh_socket->session);
@@ -303,6 +318,11 @@ RTRLIB_EXPORT int tr_ssh_init(const struct tr_ssh_config *config, struct tr_sock
 		ssh_socket->config.connect_timeout = RTRLIB_TRANSPORT_CONNECT_TIMEOUT_DEFAULT;
 	else
 		ssh_socket->config.connect_timeout = config->connect_timeout;
+
+	if (config->password)
+		ssh_socket->config.password = lrtr_strdup(config->password);
+	else
+		ssh_socket->config.password = NULL;
 
 	ssh_socket->ident = NULL;
 	ssh_socket->config.data = config->data;
--- a/rtrlib/transport/ssh/ssh_transport.c
+++ b/rtrlib/transport/ssh/ssh_transport.c
@@ -330,7 +330,6 @@ RTRLIB_EXPORT int tr_ssh_init(const struct tr_ssh_config *config, struct tr_sock
 	ssh_socket->ident = NULL;
 	ssh_socket->config.data = config->data;
 	ssh_socket->config.new_socket = config->new_socket;
-	;
 
 	return TR_SUCCESS;
 }
--- a/rtrlib/transport/ssh/ssh_transport.c
+++ b/rtrlib/transport/ssh/ssh_transport.c
@@ -256,7 +256,7 @@ int tr_ssh_recv(const void *tr_ssh_sock, void *buf, const size_t buf_len, const 
 		return TR_ERROR;
 
 	if (ssh_channel_is_eof(((struct tr_ssh_socket *)tr_ssh_sock)->channel) != 0)
-		return SSH_ERROR;
+		return TR_ERROR;
 
 	if (!rchans[0])
 		return TR_WOULDBLOCK;
@@ -267,7 +267,12 @@ int tr_ssh_recv(const void *tr_ssh_sock, void *buf, const size_t buf_len, const 
 int tr_ssh_send(const void *tr_ssh_sock, const void *pdu, const size_t len,
 		const time_t timeout __attribute__((unused)))
 {
-	return ssh_channel_write(((struct tr_ssh_socket *)tr_ssh_sock)->channel, pdu, len);
+	int ret = ssh_channel_write(((struct tr_ssh_socket *)tr_ssh_sock)->channel, pdu, len);
+
+	if (ret == SSH_ERROR)
+		return TR_ERROR;
+
+	return ret;
 }
 
 const char *tr_ssh_ident(void *tr_ssh_sock)
--- a/rtrlib/transport/ssh/ssh_transport.c
+++ b/rtrlib/transport/ssh/ssh_transport.c
@@ -290,46 +290,91 @@ RTRLIB_EXPORT int tr_ssh_init(const struct tr_ssh_config *config, struct tr_sock
 	socket->send_fp = &tr_ssh_send;
 	socket->ident_fp = &tr_ssh_ident;
 
-	socket->socket = lrtr_malloc(sizeof(struct tr_ssh_socket));
+	socket->socket = lrtr_calloc(1, sizeof(struct tr_ssh_socket));
 	struct tr_ssh_socket *ssh_socket = socket->socket;
 
 	ssh_socket->channel = NULL;
 	ssh_socket->session = NULL;
 	ssh_socket->config.host = lrtr_strdup(config->host);
+	if (!ssh_socket->config.host)
+		goto error;
 	ssh_socket->config.port = config->port;
+
 	ssh_socket->config.username = lrtr_strdup(config->username);
+	if (!ssh_socket->config.username)
+		goto error;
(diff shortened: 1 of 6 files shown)


mroethke:
ci:rerun

tools/rtrclient.c
```diff
@@ -230,6 +247,45 @@ static bool is_readable_file(const char *str)
...
+		} else if ((str[i] & UTF8_FOUR_BYTE_MASK) == UTF8_FOUR_BYTE_PREFIX && i + 3 < len &&
+			   (str[i + 1] & UTF8_SUBSEQUENT_BYTE_MASK) == UTF8_SUBSEQUENT_BYTE_PREFIX &&
+			   (str[i + 2] & UTF8_SUBSEQUENT_BYTE_MASK) == UTF8_SUBSEQUENT_BYTE_PREFIX &&
+			   (str[i + 3] & UTF8_SUBSEQUENT_BYTE_MASK) == UTF8_SUBSEQUENT_BYTE_PREFIX) {
+			i += 3;
+			continue;
+		}
+
+		// if none of the conditions matched. The string contains at least one character that are not valid utf8
+		return false;
```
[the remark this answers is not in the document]
smlng:
```
int i = 0;
bool is_utf8 = true;
while (i < len && is_utf8) {
    if (//1 byte) {
        i += 1;
    }
    else if (//2 bytes) {
        i += 2;
    }
    ...
    else {
        is_utf8 = false;
    }
}
return is_utf8;
```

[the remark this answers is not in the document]
smlng: however, just a suggestion not a blocker

tools/rtrclient.c
```diff
@@ -230,6 +247,45 @@ static bool is_readable_file(const char *str)
...
+static bool is_utf8(const char *str)
+{
+	size_t len = strlen(str);
+
+	for (size_t i = 0; i <= len; ++i) {
+		// check if current byte is a single utf char
+		if ((str[i] & UTF8_ONE_BYTE_MASK) == UTF8_ONE_BYTE_PREFIX) {
+			continue;
+
+			// check if current byte is the start of a two byte utf8 char and validate subsequent bytes
```
smlng, line 260: minor: indention of comment not aligned like comment above? Same for next 2 `else if`.

rtrlib/transport/ssh/ssh_transport.c
```diff
@@ -250,41 +250,86 @@ RTRLIB_EXPORT int tr_ssh_init(const struct tr_ssh_config *config, struct tr_sock
...
+	if (!ssh_socket->config.username)
+		goto error;
 
 	if ((config->password && config->client_privkey_path) || (!config->password && !config->client_privkey_path))
 		return TR_ERROR;
 
-	if (config->bindaddr)
+	if (config->bindaddr) {
 		ssh_socket->config.bindaddr = lrtr_strdup(config->bindaddr);
-	else
```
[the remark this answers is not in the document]
mroethke: bindaddr is optional. We must first check if it is set and than if the strdup was successful.

tools/rtrclient.c
```diff
@@ -230,6 +247,45 @@ static bool is_readable_file(const char *str)
...
+			// check if current byte is the start of a four byte utf8 char and validate subsequent bytes
+		} else if ((str[i] & UTF8_FOUR_BYTE_MASK) == UTF8_FOUR_BYTE_PREFIX && i + 3 < len &&
+			   (str[i + 1] & UTF8_SUBSEQUENT_BYTE_MASK) == UTF8_SUBSEQUENT_BYTE_PREFIX &&
+			   (str[i + 2] & UTF8_SUBSEQUENT_BYTE_MASK) == UTF8_SUBSEQUENT_BYTE_PREFIX &&
+			   (str[i + 3] & UTF8_SUBSEQUENT_BYTE_MASK) == UTF8_SUBSEQUENT_BYTE_PREFIX) {
+			i += 3;
+			continue;
+		}
+
+		// if none of the conditions matched. The string contains at least one character that are not valid utf8
```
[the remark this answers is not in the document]
mroethke: fixed

tools/rtrclient.c
```diff
@@ -230,6 +247,45 @@ static bool is_readable_file(const char *str)
 	return is_file(str);
 }
 
+#ifdef RTRLIB_HAVE_LIBSSH
+static bool is_utf8(const char *str)
+{
+	size_t len = strlen(str);
+
+	for (size_t i = 0; i <= len; ++i) {
+		// check if current byte is a single utf char
```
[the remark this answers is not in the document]
mroethke: fixed

tools/rtrclient.c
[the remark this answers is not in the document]
mroethke: I would argue that the current form is more readable, because the continue makes it clear that there is nothing relevant after the if.

tools/rtrclient.c
mroethke: fixed

tools/rtrclient.c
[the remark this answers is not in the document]
smlng: I still disagree, all `continues` are only required bc `return false` is not in an `else` branch. A curious developer will tend to look what is skipped over by `continue` anyway - and will discover its only the `return false`. I'm not against using `continue` in general, but here it seems unnecessary to me.

Further, I think it is suboptimal that `i` is manipulated twice, i.e., in the loop statement (`++i`) and by all `else if` cases. Also it is counter intuitive that `i` is increase by `byte-size - 1` - I admit that the for-loop avoids accidentally creating an invite loop, hence I still find something like

```
int i = 0;
while (i < len) {
    if (// 1 byte utf8) {
        //do stuff
        i += 1;
    }
    else if (//2 byte utf8) {
        // do stuff
        i += 2;
    }
    ...
    else {
        return false;
    }
}
return true;
```

easier to understand and less error prone.

tools/rtrclient.c
```diff
@@ -230,6 +247,45 @@ static bool is_readable_file(const char *str)
 	return is_file(str);
 }
 
+#ifdef RTRLIB_HAVE_LIBSSH
+static bool is_utf8(const char *str)
+{
+	size_t len = strlen(str);
+
+	for (size_t i = 0; i <= len; ++i) {
```
smlng, line 255: shouldn't this be `i < len` instead of `i <= len`?

Arvedui: Yes. Though in practice i does not matter because the last byte is always the terminating null byte.
I fixed it anyway of course.

tools/rtrclient.c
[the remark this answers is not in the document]
Arvedui: I'm starting to think that you are right. Change it accordingly.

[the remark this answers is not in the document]
mroethke: I'm starting to think that you are right, changed it accordingly.

tools/rtrclient.c
mroethke: Yes. fixed.

mroethke:
I also noticed two additional problems myself.
1. I forgot to amend the help string and the man page
2. The option `-s` to force the rtrclient to interpret the string as a path to a private key collided with another global option. I switched it to use `-r`.

mroethke:
ping @pguibert6WIND

rtrlib/transport/ssh/ssh_transport.c
```diff
@@ -137,14 +137,29 @@ int tr_ssh_open(void *socket)
 		goto error;
 	}
 
+	if (config->client_privkey_path) {
+		SSH_DBG("%s: Trying publickey authentication", ssh_socket, __func__);
+		ssh_options_set(ssh_socket->session, SSH_OPTIONS_IDENTITY, config->client_privkey_path);
+
+		int rtval;
+
```
pguibert6WIND, line 149: is it possible to swap the order:
'int rtval'
before ssh_options_set, 

like most of C programs ?

rtrlib/transport/ssh/ssh_transport.c
```diff
@@ -304,6 +319,11 @@ RTRLIB_EXPORT int tr_ssh_init(const struct tr_ssh_config *config, struct tr_sock
 	else
 		ssh_socket->config.connect_timeout = config->connect_timeout;
 
+	if (config->password)
+		ssh_socket->config.password = lrtr_strdup(config->password);
+	else
```
pguibert6WIND: how is it freed  ?

cmake/modules/FindLibSSH.cmake
```diff
@@ -73,6 +73,10 @@ else (LIBSSH_LIBRARIES AND LIBSSH_INCLUDE_DIRS)
                 string(REGEX MATCH "[0-9]+" LIBSSH_VERSION_PATCH ${LIBSSH_VERSION_PATCH})
 
                 set(LibSSH_VERSION ${LIBSSH_VERSION_MAJOR}.${LIBSSH_VERSION_MINOR}.${LIBSSH_VERSION_PATCH})
+                set(LibSSH_VERSION_MAJOR ${LIBSSH_VERSION_MAJOR})
+                set(LibSSH_VERSION_MINOR ${LIBSSH_VERSION_MINOR})
+                set(LibSSH_VERSION_PATCH ${LIBSSH_VERSION_PATCH})
+                set(LibSSH_VERSION_COUNT "3")
```
pguibert6WIND, line 80: why 3?

rtrlib/transport/ssh/ssh_transport.c
```diff
@@ -299,6 +299,9 @@ RTRLIB_EXPORT int tr_ssh_init(const struct tr_ssh_config *config, struct tr_sock
 	ssh_socket->config.port = config->port;
 	ssh_socket->config.username = lrtr_strdup(config->username);
 
+	if ((config->password && config->client_privkey_path) || (!config->password && !config->client_privkey_path))
+		return TR_ERROR;
+
```
pguibert6WIND, line 321: could not we set  a priority ? first privkey, trhen password if privkey fails ?

rtrlib/transport/ssh/ssh_transport.c
```diff
@@ -290,46 +290,91 @@ RTRLIB_EXPORT int tr_ssh_init(const struct tr_ssh_config *config, struct tr_sock
...
 	ssh_socket->channel = NULL;
 	ssh_socket->session = NULL;
 	ssh_socket->config.host = lrtr_strdup(config->host);
+	if (!ssh_socket->config.host)
+		goto error;
 	ssh_socket->config.port = config->port;
+
 	ssh_socket->config.username = lrtr_strdup(config->username);
+	if (!ssh_socket->config.username)
+		goto error;
```
pguibert6WIND, line 318: move it at same level as config.host. it avoids strdupping (config->username) for notthing.

rtrlib/transport/ssh/ssh_transport.c
```diff
@@ -243,11 +243,10 @@ int tr_ssh_recv_async(const struct tr_ssh_socket *tr_ssh_sock, void *buf, const
 	return rtval;
 }
```
pguibert6WIND, line 245: is commit log consistent with modified code ? (ssh_send and tr_ssh recv ?)

pguibert6WIND (changes requested):

rtrlib/transport/ssh/ssh_transport.c
mroethke: I prefer to declare variables were they are first used not at the beginning at the block. The declaration is only "freestanding" because it wanted to avoid duplication the declaration in the following if/else macro.

rtrlib/transport/ssh/ssh_transport.c
mroethke: it is freed in `tr_ssh_free`

cmake/modules/FindLibSSH.cmake
mroethke: Because the version number of libssh has 3 components. It a standard variable that is expected to be exported by cmakd find scripts.

rtrlib/transport/ssh/ssh_transport.c
mroethke: I don't think that makes sense in non interactive environments.

rtrlib/transport/ssh/ssh_transport.c
mroethke: You are saying I should check if username is set before strduping it?

rtrlib/transport/ssh/ssh_transport.c
mroethke: fixed in my local branch

mroethke:
ping @pguibert6WIND

waehlisch:
ci:rerun

PR #257, opened 2020-06-18
Fix rtr_stop blocking behavior of tr_open

### LINKED ISSUE #254: RTRlib/ some delay when stopping a RPKI connection with FRR
### Description
When using with FRR, BGP RPKI sometimes freezes with SSH configuration

### reproduction
I did it with [0]
with a RPKI configuration that we delete, and we recreate a new one. we can observe a 12 second delay between the calls to pthread_cancel() and pthread_join()
I do it with a ssh config, but I dont need any remote ssh server to do  that.

[0] https://github.com/FRRouting/frr/pull/5015

note: the problem happens when using 22 port, but not when using != 22 port.
### Expected results
```
Mar 03 15:24:02 bgpd[31859]: (2020/03/03 15:24:02:819582): RTR Socket: pthread_cancel()
Mar 03 15:24:02  bgpd[31859]: (2020/03/03 15:24:02:819600): RTR Socket: pthread_join()
Mar 03 15:24:02  bgpd[31859]: (2020/03/03 15:25:14:935700): SSH Transport(abcd@1.0.0.1:22): tr_ssh_init: opening SSH connection failed
```

### Actual results
```
Mar 03 15:24:02 bgpd[31859]: (2020/03/03 15:24:02:819582): RTR Socket: pthread_cancel()
Mar 03 15:24:02  bgpd[31859]: (2020/03/03 15:24:02:819600): RTR Socket: pthread_join()
Mar 03 15:25:14  bgpd[31859]: (2020/03/03 15:25:14:935700): SSH Transport(abcd@1.0.0.1:22): tr_ssh_init: opening SSH connection failed
```
### Analysis
it looks like in rtr.c file, we call `pthread_setcancelstate(PTHREAD_CANCEL_DISABLE, &oldcancelstate);` in the pthread created. then I guess the following calls will be kind of blocked.
```
rtr.c:          RTR_DBG1("pthread_cancel()");
rtr.c:          pthread_cancel(rtr_socket->thread_id);
rtr.c:          RTR_DBG1("pthread_join()");
rtr.c:          pthread_join(rtr_socket->thread_id, NULL);
```

mroethke:
### Contribution description
Since the move from pthread_kill to pthread_cancel a rtr_socket could
not be stopped immediately if it was in the tr_open call.
cancelability could not be enabled for that function because it holds
local resources.
This enables cancalability locally for the only blocking call in
tr_tcp_open and tr_ssh_open respectively.

### Testing procedure

Call `rtr_mgr_stop` shortly after `rtr_mgr_start` and make sure that the connection attempt does not succeed or fail immediately. For instance by dropping the packets via the firewall.

Without this patch `rtr_mgr_stop` hangs until the connection attempts times out.

### Issues/PRs references
Fixes #254

### DIFF
--- a/rtrlib/transport/ssh/ssh_transport.c
+++ b/rtrlib/transport/ssh/ssh_transport.c
@@ -105,8 +105,7 @@ int tr_ssh_open(void *socket)
 
 			FD_ZERO(&rfds);
 			FD_SET(fd, &rfds);
-			// timeout of 30 seconds
-			struct timeval timeout = { .tv_sec = 30, .tv_usec = 0 };
+			struct timeval timeout = {.tv_sec = ssh_socket->config.connect_timeout, .tv_usec = 0};
 			int oldcancelstate;
 
 			/* Enable cancellability for the select call
@@ -299,6 +298,11 @@ RTRLIB_EXPORT int tr_ssh_init(const struct tr_ssh_config *config, struct tr_sock
 		ssh_socket->config.server_hostkey_path = lrtr_strdup(config->server_hostkey_path);
 	else
 		ssh_socket->config.server_hostkey_path = NULL;
+
+	if (config->connect_timeout == 0)
+		ssh_socket->config.connect_timeout = RTRLIB_TRANSPORT_CONNECT_TIMEOUT_DEFAULT;
+	else
+		ssh_socket->config.connect_timeout = config->connect_timeout;
 
 	ssh_socket->ident = NULL;
 	ssh_socket->config.data = config->data;
--- a/rtrlib/transport/ssh/ssh_transport.c
+++ b/rtrlib/transport/ssh/ssh_transport.c
@@ -16,7 +16,9 @@
 #include "rtrlib/transport/transport_private.h"
 
 #include <assert.h>
+#include <errno.h>
 #include <libssh/libssh.h>
+#include <pthread.h>
 #include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -46,6 +48,7 @@ static int tr_ssh_send(const void *tr_ssh_sock, const void *pdu, const size_t le
 static int tr_ssh_recv_async(const struct tr_ssh_socket *tr_ssh_sock, void *buf, const size_t buf_len);
 static const char *tr_ssh_ident(void *tr_ssh_sock);
 
+/* WARNING: This function has cancelable sections! */
 int tr_ssh_open(void *socket)
 {
 	struct tr_ssh_socket *ssh_socket = socket;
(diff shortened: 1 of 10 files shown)


rtrlib/transport/ssh/ssh_transport.c
```diff
@@ -86,10 +89,47 @@ int tr_ssh_open(void *socket)
...
+			SSH_DBG1("tr_ssh_init: opening SSH connection failed", ssh_socket);
+			goto error;
+		} else if (ret == SSH_AGAIN) {
+			socket_t fd = ssh_get_fd(ssh_socket->session);
+
+			fd_set rfds;
+
+			FD_ZERO(&rfds);
+			FD_SET(fd, &rfds);
+			struct timeval tv = { 30, 0 };
```
smlng, line 108: add comment that this is a timeout of 30s for `select` below. Also maybe change variable name to `timeout`?

rtrlib/transport/tcp/tcp_transport.c
```diff
@@ -91,13 +136,65 @@ int tr_tcp_open(void *tr_socket)
...
+		pthread_setcancelstate(PTHREAD_CANCEL_ENABLE, &oldcancelstate);
+		int ret = select(tcp_socket->socket + 1, NULL, &wfds, NULL, NULL);
+
+		pthread_setcancelstate(PTHREAD_CANCEL_DISABLE, &oldcancelstate);
+
+		if (ret < 0) {
+			TCP_DBG("Could not select tcp socket, %s", tcp_socket, strerror(errno));
+			goto end;
+
+		} else if (ret == 0) {
```
smlng, line 178: can this actually happen, as there is no timeout given in `select` above?

smlng (commented):
minor stuff

rtrlib/transport/tcp/tcp_transport.c
mroethke: I'm wondering why I used select without a timeout there. I do not remember a specific reason and it does seem wrong to me right now.
I think I will just add a timeout like I have in the ssh implementation.

rtrlib/transport/ssh/ssh_transport.c
```diff
@@ -86,10 +89,47 @@ int tr_ssh_open(void *socket)
...
+			int oldcancelstate;
+
+			/* Enable cancalability for the select call
+			 * to prevent rtr_stop from blocking for a long time.
```
pguibert6WIND, line 112: typo: cancellability

rtrlib/transport/tcp/tcp_transport.c
```diff
@@ -91,13 +136,65 @@ int tr_tcp_open(void *tr_socket)
...
+		res = NULL;
+
+		fd_set wfds;
+
+		FD_ZERO(&wfds);
+		FD_SET(tcp_socket->socket, &wfds);
+		int oldcancelstate;
+
+		/* Enable cancalability for the select call
+		 * to prevent rtr_stop from blocking for a long time.
```
pguibert6WIND, line 165: typo: cancelability

rtrlib/transport/ssh/ssh_transport.c
pguibert6WIND: > add comment that this is a timeout of 30s for `select` below. Also maybe change variable name to `timeout`?

+1. agreed. by the way, having the timeout value configurable would be a nice to have. ( separate ticket, or separate commit )

pguibert6WIND (changes requested):
actually, select() is used because timeout param can be used.
if you had not the timeout to handle, you could have used read() to simplify the code.

actually, select() is mandatory with timeout.

rtrlib/transport/ssh/ssh_transport.c
mroethke: > add comment that this is a timeout of 30s for `select` below. Also maybe change variable name to `timeout`?

done


> > add comment that this is a timeout of 30s for `select` below. Also maybe change variable name to `timeout`?
> 
> +1. agreed. by the way, having the timeout value configurable would be a nice to have. ( separate ticket, or separate commit )

Yea, making this configurable is probably a good idea. Will add this to this PR.

rtrlib/transport/ssh/ssh_transport.c
mroethke: done

rtrlib/transport/ssh/ssh_transport.h
```diff
@@ -45,6 +45,7 @@
  *	  is made. The returned socket is expected to be ready for use (e.g.
  *	  in state established), and must use a reliably stream-oriented transport.
  *	  When new_socket() is used, host, port, and bindaddr are not used.
+ * @param connect_timeout Time in seconds to wait for a succeccfull connection. Defaults to 30 seconds
  */
```
pguibert6WIND, line 49: nit: s/succeccfull/successfull

rtrlib/transport/tcp/tcp_transport.h
```diff
@@ -35,13 +35,15 @@
  *	  is made. The returned socket is expected to be ready for use (e.g.
  *	  in state established), and must use a reliably stream-oriented transport.
  *	  When new_socket() is used, host, port, and bindaddr are not used.
+ * @param connect_timeout Time in seconds to wait for a succeccfull connection. Defaults to 30 seconds
  */
```
pguibert6WIND, line 39: s/succeccfull/successfull

pguibert6WIND (changes requested):
2 typo comments.

I hesitate to ask you to move TIMEOUT define value to transport header.
or to add on doxygen generated documentation a seealso information so that people can easily identify the timeout by default.

as for the latter, I don't know how to do it with doxygen, and the first is self contained, I think it is ok like that.

thanks.

rtrlib/transport/ssh/ssh_transport.h
mroethke: actually it is [sucessful](https://www.merriam-webster.com/dictionary/successful), but it is wrong either way, so thank for pointing that out.

mroethke:
ci:rerun

mroethke:
ci:rerun

mroethke:
> I hesitate to ask you to move TIMEOUT define value to transport header.
> or to add on doxygen generated documentation a seealso information so that people can easily identify the timeout by default.
> 
> as for the latter, I don't know how to do it with doxygen, and the first is self contained, I think it is ok like that.

I moved the connect timeout to the transport header and referenced the define instead of writing out the timeout. I think that is the best we can do with doxygen.

pguibert6WIND:
Hi Marcel,
I would be ok to approve the pr, but first, could you clarify me why codecov/patch and codecov/project fails ?
if this is unrelated, then I will approve your pr.
thanks,

mroethke:
Frankly codecov is pretty broken, at least for us. Most of the lines that are not covered anymore did not change and changes made should not have an impact. I'm not sure why it is so broken, but it unfortunately is. I think it is still nice as ballpark metric, but should not be used to block pull request except in very extreme cases.

pguibert6WIND (approved):

PR #258, opened 2020-06-24
rtrlib/transport/tcp: add extended errno output

colinbs:
### Contribution description

In case `tr_tcp_open` receives an _EAI_SYSTEM_ error code, `gai_strerror` would only print "Unknown error". `strerror` will then print a more detailed error string, e.g. "Resource temporarily unavailable". This error string is more useful for debugging purposes.

### DIFF
--- a/rtrlib/transport/tcp/tcp_transport.c
+++ b/rtrlib/transport/tcp/tcp_transport.c
@@ -47,6 +47,7 @@ static const char *tr_tcp_ident(void *socket);
 int tr_tcp_open(void *tr_socket)
 {
 	int rtval = TR_ERROR;
+	int tcp_rtval = 0;
 	struct tr_tcp_socket *tcp_socket = tr_socket;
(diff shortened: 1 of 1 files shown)


mroethke:
I'm not sure this is correct.

As far as I understand from the man page, `errno` should not be passed to `gai_strerror` at all. The return value of `getaddrinfo` should be. Only if the return value is equal to `EAI_SYSTEM` should errno be checked.

colinbs:
> I'm not sure this is correct.
> 
> As far as I understand from the man page, `errno` should not be passed to `gai_strerror` at all. The return value of `getaddrinfo` should be. Only if the return value is equal to `EAI_SYSTEM` should errno be checked.

Ah yes, I see. I changed it to using a designated return variable for the calls. On all errors except `EAI_SYSTEM`, the variable `tcp_rtval` will be used with `gai_strerror` now.
I noticed quite a new other places where `errno` is passed to `gai_strerror`. It would be convenient to change these occurences aswell, don't you think?

PR #259, opened 2020-09-16
tests: fix warning about discarded qualifier in live validation test

mroethke:
### Contribution description

This fixes a build error on freebsd11.

This should probably be merged before #257  and #256

### DIFF
--- a/tests/test_live_validation.c
+++ b/tests/test_live_validation.c
@@ -21,9 +21,6 @@ struct test_validity_query {
 	unsigned int val;
 };
 
-static char *RPKI_CACHE_HOST = "rpki-validator.realmv6.org";
-static char *RPKI_CACHE_POST = "8283";
-
 /*
  * Verification is based on ROAs for RIPE RIS Routing Beacons, see:
  * (https://www.ripe.net/analyse/internet-measurements/
@@ -57,6 +54,12 @@ static void connection_status_callback(const struct rtr_mgr_group *group __attri
  */
 int main(void)
 {
+	/* These variables are not in the global scope
+	 * because it would cause warnings about discarding constness
+	 */
+	char RPKI_CACHE_HOST[] = "rpki-validator.realmv6.org";
+	char RPKI_CACHE_POST[] = "8283";
+
 	/* create a TCP transport socket */
 	struct tr_socket tr_tcp;
 	struct tr_tcp_config tcp_config = {RPKI_CACHE_HOST, RPKI_CACHE_POST, NULL, NULL, NULL};


tests/test_live_validation.c
```diff
@@ -57,6 +54,12 @@ static void connection_status_callback(const struct rtr_mgr_group *group __attri
  */
 int main(void)
 {
+	/* These variables are not in the global scope
+	 * because it would cause warnings about discarding constness
+	 */
+	char RPKI_CACHE_HOST[] = "rpki-validator.realmv6.org";
+	char RPKI_CACHE_POST[] = "8283";
+
```
pguibert6WIND, line 62: ok, I understand this avoids casting .
I will approve the change.
I just need to understand how come test report is wrong?

pguibert6WIND (changes requested):

tests/test_live_validation.c
mroethke: [This](https://ci1.netdef.org/browse/RPKI-RTRLIBPR-FBSD11AMD64-309/log) build log illustrates the problem. Towards the bottom above `make (Debug) failed`. 

```
22-Sep-2020 02:54:53 	/usr/home/ci/cibuild.309/rtrlib-source/tests/test_live_validation.c:24:14: error: initializing 'char *' with an expression of type 'const char [27]' discards qualifiers [-Werror,-Wincompatible-pointer-types-discards-qualifiers]
22-Sep-2020 02:54:53 	static char *RPKI_CACHE_HOST = "rpki-validator.realmv6.org";
22-Sep-2020 02:54:53 	             ^                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
22-Sep-2020 02:54:53 	/usr/home/ci/cibuild.309/rtrlib-source/tests/test_live_validation.c:25:14: error: initializing 'char *' with an expression of type 'const char [5]' discards qualifiers [-Werror,-Wincompatible-pointer-types-discards-qualifiers]
22-Sep-2020 02:54:53 	static char *RPKI_CACHE_POST = "8283";
22-Sep-2020 02:54:53 	             ^                 ~~~~~~
```

Does that answer you question?

pguibert6WIND: partially. yes; your patch is worth pusing it.
actually I think it is the same with all builds: codecov/project fails.
but this is due to test coverage not enough, and should be part of an other ticket.

pguibert6WIND (approved):

tests/test_live_validation.c
mroethke: Ah now I get what you meant. It is actually not because test coverage is to low, it is because codecov thins the coverage went down with this PR. Which does not make sense as it does not touch any c code. 

Thats something we should probably investigate at some point.

mroethke:
Thanks for your review!

PR #260, opened 2020-09-22
cmake: use FindPkgConfig to search for libssh availability

mroethke:
### Contribution description
Previously a homegrown find script was used. That made assumptions about locations of some header files that do not hold true anymore for recent libssh versions. The canonical why of finding libraries is pkg-config and cmake has excellent support for that. So we simply use it.


colinbs (approved):

mroethke:
Thanks for the review!

PR #268, opened 2021-12-09
rtrlib/rtr_mgr: properly cleanup rtr_sockets on stop

### LINKED ISSUE #267: Group Shutdown
Hi, while using the example I notice that when the most preferable group had a connection error,
and the second group start and connect, when the most preferable group connection restored the second group
is shutdown and cannot start again.

I look on the code and notice that you prevent from the group to run again after shutdown:

void *rtr_fsm_start(struct rtr_socket *rtr_socket)
{
    if (rtr_socket->state == RTR_SHUTDOWN)
        return NULL;
...

Why when group connected and shutdown it cannot start again?

Yakov S.

mroethke:
Previously rtr_sockets could not be restarted because their state
remained on SHUTDOWN, which they can, by design, not recover from
themselves.

fix #267

### DIFF
--- a/rtrlib/rtr/rtr.c
+++ b/rtrlib/rtr/rtr.c
@@ -241,6 +241,7 @@ void rtr_stop(struct rtr_socket *rtr_socket)
 		pfx_table_src_remove(rtr_socket->pfx_table, rtr_socket);
 		spki_table_src_remove(rtr_socket->spki_table, rtr_socket);
 		rtr_socket->thread_id = 0;
+		rtr_socket->state = RTR_CLOSED;
 	}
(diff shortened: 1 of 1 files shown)


waehlisch:
@Jaykov-s can you check that this fixes your problem?

Jaykov-s:
Hi @waehlisch Sorry for the late reply, I was assigned to another task.
I tested your solution and it solved the problem.
Thank you,

Regards, Yakov S.

waehlisch:
thanks @Jaykov-s and @mroethke!

waehlisch (approved):

PR #270, opened 2022-01-05
rtr_mgr: replace mutex with rwlock

### LINKED ISSUE #269: Mutex Deadlock Issue
Hi, I found issues while running session with a multiple groups, when connections is down for the most preferable group,
the second preferable group started and connected. When the most preferable group connection restored, the less preferable
group session is closed and the system continue working with the most preferable group. I found that in some cases,
when the connection is down, and restored, Or the whole session is terminated by running "rtr_mgr_stop()" we got Mutex deadlock.

I create a session with the follow 8 groups:
	1. existing local server address: 10.1.68.191, port: 8282, preference: 1
	2. existing local server address: 10.1.68.134, port: 8282, preference: 2
	3. dummy server address: 1.1.1.1, port 1111, preference: 3
	4. dummy server address: 2.2.2.2, port 2222, preference: 4
	5. dummy server address: 3.3.3.3, port 3333, preference: 5
	6. dummy server address: 4.4.4.4, port 4444, preference: 6
	7. dummy server address: 5.5.5.5, port 5555, preference: 7
	8. dummy server address: 6.6.6.6, port 6666, preference: 8
		
The reason that I add more dummy servers is to cause the issue appear more often.
I added more prints in the code for Mutex-Lock, Mutex-unlock operations.

In the Log file that include, we can see the scenario that cause the system to be in mutex deadlock, In this case,
Server 1 and 2 was up and then both server 1 and 2 was down and restored, I added the follow comments to the log file (mutex-deadlock-issue-remarks.log)

1.	Thread 1 - Group 1 Connection establish, Close all the other Groups

2.	Thread 1 - Close Group 2 Start,

3.	Thread 7 - Group 7 Attempt to connect and failed, Run: "is_some_rtr_mgr_group_established" and Lock the Mutex.
	Thread 7 now is sleep until the Mutex is unlock.

4.	Thread 8 - Group 8 Attempt to connect and failed same as group 7, Run: "is_some_rtr_mgr_group_established" and Lock the 
	Mutex. Thread 8 now is sleep until the Mutex is unlock.
	
5.	Thread 1 - Continue closing group 2

6.	Thread 1 - Group 2 is successfully closed, Closing Group 3.

7.	Thread 1 - Group 3 is successfully closed, Closing Group 4.

8.	Thread 1 - Group 4 is successfully closed, Closing Group 5.

9.	Thread 1 - Group 5 is successfully closed, Closing Group 6.

10.	Thread 1 - Group 6 is successfully closed, Closing Group 7.

11.	Thread 1 - Running "pthread_cancel" and "pthread_join" for thread 7 (group 7),
	As it normal operation, thread 7 main state machine should reach the: RTR_SHUTDOWN state but thread 7 is sleep
	because the Mutex locked by thread 1, So thread 1 waiting on pthread_join, while thread 7 is sleep because the Mutex.
	
12.  The system is in deadlock.

I attached the follow files:
1.	rtr_static.c - Our example with all the server above.
2. 	rtr_mgr.c - Modified file with the Mutex Lock/Unlock Debug Prints.
3. 	mutex-deadlock-issue.log - original log.
4. 	mutex-deadlock-issue-remarks.log - Log with the above remarks, (starting with symble: >>+)

Can you verify this issue,
Regards, Yakov S.

[mutex-deadlock-issue.log](https://github.com/rtrlib/rtrlib/files/7781126/mutex-deadlock-issue.log)
[mutex-deadlock-issue-remarks.log](https://github.com/rtrlib/rtrlib/files/7781127/mutex-deadlock-issue-remarks.log)
[rtr_mgr.c.txt](https://github.com/rtrlib/rtrlib/files/7781128/rtr_mgr.c.txt)
[rtr_static.c.txt](https://github.com/rtrlib/rtrlib/files/7781129/rtr_static.c.txt)

mroethke:
rtr_mgr could enter a deadlocked state with multiple socket groups when
a group with at least one group of lower preference comes back online.
This happens because the thread of the group coming back online blocks
on trying to shut down all threads with lower preference while holding
the rtr_mgrs mutex, but a thread that tries to acquire that same mutex
cannot be shut down while doing so.

Since socket threads do not change rtr_mgrs protected structures an
rwlock can be used to allow multiple socket threads to read them while
preventing external calls from changing then concurrently. This allows
the thread that is supposed to shut down to enter code sections where it
can be shut down.

Fix #269

### DIFF
--- a/rtrlib/rtr_mgr.c
+++ b/rtrlib/rtr_mgr.c
@@ -189,7 +189,7 @@ static inline void _rtr_mgr_cb_state_established(const struct rtr_socket *sock, 
 		 */
 		bool all_error = true;
 
-		pthread_rwlock_wrlock(&config->mutex);
+		pthread_rwlock_rdlock(&config->mutex);
 		tommy_node *node = tommy_list_head(&config->groups->list);
 
 		while (node) {
--- a/rtrlib/rtr_mgr.c
+++ b/rtrlib/rtr_mgr.c
@@ -94,7 +94,7 @@ bool rtr_mgr_config_status_is_synced(const struct rtr_mgr_group *group)
 static void rtr_mgr_close_less_preferable_groups(const struct rtr_socket *sock, struct rtr_mgr_config *config,
 						 struct rtr_mgr_group *group)
 {
-	pthread_mutex_lock(&config->mutex);
+	pthread_rwlock_rdlock(&config->mutex);
 	tommy_node *node = tommy_list_head(&config->groups->list);
 
 	while (node) {
@@ -109,12 +109,12 @@ static void rtr_mgr_close_less_preferable_groups(const struct rtr_socket *sock, 
 		}
 		node = node->next;
 	}
-	pthread_mutex_unlock(&config->mutex);
+	pthread_rwlock_unlock(&config->mutex);
 }
 
 static struct rtr_mgr_group *get_best_inactive_rtr_mgr_group(struct rtr_mgr_config *config, struct rtr_mgr_group *group)
 {
-	pthread_mutex_lock(&config->mutex);
+	pthread_rwlock_rdlock(&config->mutex);
 	tommy_node *node = tommy_list_head(&config->groups->list);
 
 	while (node) {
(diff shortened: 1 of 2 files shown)


mroethke:
@Jaykov-s please confirm that this fixes the bug

Jaykov-s:
@mroethke I'm running test on your solution, I see on line: 202 that you use: pthread_rwlock_wrlock, is this correct? should we need to use: pthread_rwlock_rdlock in this section?
Jaykovs

mroethke:
line 202 of what? There is no function call in `rtr_mgr.c` on that line.

Do you mean the one on line 192? That one should indeed be a call to `pthread_rwlock_rdlock` instead.
I have pushed a fix.

mwinter-osr:
CI:rerun

waehlisch (approved):
works.

PR #272, opened 2022-01-17
redhat: Fix RPM file to work on Fedora 33+ and RedHat 9+

mwinter-osr:
- Newer Fedora/RedHat changes cmake rpm build to build out of tree
  and requires the use of macros for correct locations
  See https://fedoraproject.org/wiki/Changes/CMake_to_do_out-of-source_builds
- On newer Fedora, the SOURCES subdirectory isn't created
  automatically and needs to be created in the prep phase before
  the tar is created

Signed-off-by: Martin Winter <mwinter@opensourcerouting.org>


mroethke (approved):
Thanks!

PR #277, opened 2022-09-22
Restrict overmatching MACH ifdef to only trigger on OSX and Mach

Zopolis4:
### Contribution description

Hurd also uses Mach.
https://www.gnu.org/software/hurd/hurd/porting/guidelines.html
```
#ifdef __MACH__
Some applications put Apple Darwin-specific code inside #ifdef __MACH__ guards. Such guard is clearly not enough, since not only Apple uses Mach as a kernel. This should be replaced by #if defined(__MACH__) && defined(__APPLE__)
```

### Testing procedure

Try to compile rtrlib on Hurd.

### Issues/PRs references

N/A

### DIFF
--- a/rtrlib/lib/utils.c
+++ b/rtrlib/lib/utils.c
@@ -13,14 +13,14 @@
 #include <stdint.h>
 #include <time.h>
 
-#ifdef __MACH__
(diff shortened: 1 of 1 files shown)


mroethke (approved):
looks good!
Thanks!

PR #279, opened 2023-10-20
[DRAFT] Vector Data-Structure

tanneberger:
### Contribution 

- [ ] ASPA Data Structure PR: https://github.com/revol-xut/rtrlib/pull/1

### DIFF
--- a/rtrlib/lib/ordered_dyn_array.h
+++ b/rtrlib/lib/ordered_dyn_array.h
@@ -1,0 +1,96 @@
+/*
+* This file is part of RTRlib.
+*
+* This file is subject to the terms and conditions of the MIT license.
+* See the file LICENSE in the top level directory for more details.
+*
+* Website: http://rtrlib.realmv6.org/
+*/
+
+#ifndef LRTR_VECTOR_UTILS_H
+#define LRTR_VECTOR_UTILS_H
+
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+
+/**
+ * @brief ASPA Object
+ * here the customer as number (CAS) authorizes a set of provider AS numbers
+ */
+struct aspas_record {
+	uint32_t customer_asn;
+	size_t provider_count;
+	uint32_t *provider_asns;
+};
+
+/**
+ * @brief Struct which is similar in function to std::vector from C++.
+ * If the vector is running full a larger chunk of memory is allocated and the data is copied over.
+ */
+struct ordered_dyn_array {
+	uint32_t size;
+	size_t capacity;
+	struct aspas_record *data;
+};
+/**
+ * creates an vector object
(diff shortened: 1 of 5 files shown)


waehlisch:
@mroethke @fho any feedback is highly appreciated. a group of students at TUD started to integrate [ASPA](https://datatracker.ietf.org/doc/html/draft-ietf-sidrops-aspa-verification) into RTRlib.

revol-xut:
Please note that the main development will first happen in the fork https://revol-xut/rtrlib, and then we will do a larger PR that adds full ASPA support at once, following the rfc @waehlisch just posted. 

So this PR will grow slowly over the next couple of weeks.

mroethke:
Several quick comments:
- Commit messages should be in imparative mood `lib: add aspa_vector …`
- Please also explain the why in the commit message. Why is this data structure needed? Why this kind of data structure and not something else. 
- I don't really like the name, it is an ordered dynamic array, please call it that. (ordered_dyn_array or something like that would also work for me)
- Please use the malloc wrapper from `rtrlib/lib/alloc_utils.h` instead of calling malloc and friends directly.

I have not really looked at the code so far, I will do that after you have explained the why.

rtrlib/lib/ordered_dyn_array.h
```diff
@@ -0,0 +1,96 @@
...
+ * here the customer as number (CAS) authorizes a set of provider AS numbers
+ */
+struct aspas_record {
+	uint32_t customer_asn;
+	size_t provider_count;
+	uint32_t *provider_asns;
+};
+
+/**
+ * @brief Struct which is similar in function to std::vector from C++.
```
fho, line 28: Instead of referencing std::vector here, it would be better to describe what it does.
Otherwise the reader first have to figure out how std::vector functions to know what ordered_dyn_array does

fho (commented):
great project, the code looks good, I reviewed ~80. 

1. I'm wondering why it is chosen to dynamically resize the array.
   Why does it fit better the user case then then letting the caller specify the
   initial size + providing a resize operation where the new size is explicitly
   passed?
   The reason should also be documented in the commit and/or PR description.
2. It's not needed to have a comment for every line.
   Most of the  the code is self-explanatory, e.g. when calling `free` or assigning an array element. Having comments for those simple operations makes it actually more cumbersome to read.
3. The naming became inconsistent. In some places the data structure is called
   vector (var names, documentations) in other places ordered_dyn_array.
4. Use more error codes then -1.
   Currently only -1 is returned on errors. This makes it  very hard to impossible for the  caller to figure why an operation failed.

fho:
@tanneberger can this PR be closed? Seems to be superseded by https://github.com/rtrlib/rtrlib/pull/285

fho:
> @tanneberger can this PR be closed? Seems to be superseded by #285

I assume, no reply means "Yes".
Feel free to reopen the PR otherwise.

PR #288, opened 2024-04-05
Bugfix: Use the actual struct size and not the pointer size

nbars:


tanneberger (approved):
Good Catch

Precisely the reason why I prefer `sizeof(actual_data_type)` over `sizeof(variable)`.

Nimuedaertya (commented):
Preferring as well the solution `sizeof(trie_node)`

waehlisch:
@nbars, maybe you can comment why you finally decided on `now: sizeof(*n)` instead of the preferences of @tanneberger and @Nimuedaertya ... in case questions pop up later.

nbars:
Sure; The rtrlib is (or should be) formatted according to the kernel coding style. And the style mandates to use the `sizeof(*n)` notation.

PR #298, opened 2025-06-16
Fix: Add missing memory allocation NULL-checks

### LINKED ISSUE #297: lrtr_malloc without NULL check
### Description
In /rtrlib/rtrlib/transport/tcp/tcp_transport.c, after calling the` lrtr_malloc` function on line 341, no check is performed for the returned NULL value. However, the returned value is usually checked in multiple other cases. The returned value is then dereferenced on line 344.

### Expected results
```
	socket->socket = lrtr_malloc(sizeof(struct tr_tcp_socket));
	struct tr_tcp_socket *tcp_socket = socket->socket;

        if (tcp_socket) {
	    tcp_socket->socket = -1;
```

### Actual results
```
	socket->socket = lrtr_malloc(sizeof(struct tr_tcp_socket));
	struct tr_tcp_socket *tcp_socket = socket->socket;

	tcp_socket->socket = -1;
```

Brias:
### Contribution description

#297 reports a missing NULL-check for a memory allocation in `/rtrlib/rtrlib/transport/tcp/tcp_transport.c` which results in undefined behavior / a segmentation fault due to dereferencing a NULL-pointer if the memory allocation failed.

After searching for the memory allocation functions `lrtr_malloc`, `lrtr_calloc`, and `lrtr_realloc` in the source code (i.e. within the `rtrlib/` directory) I found multiple other locations in the code (i.e. in the modules `bgpsec`, `aspa`, and `transport`) where the NULL-check was missing.

This pull request adds the missing NULL-checks, frees previously allocated data if necessary and returns an error code/`NULL` in case the memory allocation failed. In addition, it fixes a memory leak in `tr_ssh_init` which was caused by a premature return when either both the SSH password and path to the private key or none of them are configured.

### Testing procedure

n/a

### Issues/PRs references

Fixes #297

### DIFF
--- a/rtrlib/aspa/aspa.c
+++ b/rtrlib/aspa/aspa.c
@@ -32,6 +32,11 @@ static enum aspa_status aspa_table_notify_clients(struct aspa_table *aspa_table,
 			size_t size = sizeof(uint32_t) * record->provider_count;
 
 			rec.provider_asns = lrtr_malloc(size);
+
+			if (rec.provider_asns == NULL) {
+				return ASPA_ERROR;
+			}
+
 			memcpy(rec.provider_asns, record->provider_asns, size);
 		} else {
 			rec.provider_asns = NULL;
--- a/rtrlib/bgpsec/bgpsec_utils.c
+++ b/rtrlib/bgpsec/bgpsec_utils.c
@@ -25,30 +25,34 @@ struct stream {
 
 struct stream *init_stream(uint16_t size)
 {
-	struct stream *s = lrtr_calloc(sizeof(struct stream), 1);
+	struct stream *stream = lrtr_calloc(sizeof(struct stream), 1);
 
-	if (!s) {
+	if (stream == NULL) {
 		return NULL;
 	}
 
-	s->stream = lrtr_calloc(size, 1);
+	stream->stream = lrtr_calloc(size, 1);
 
-	if (!s->stream) {
-		lrtr_free(s);
+	if (stream->stream == NULL) {
+		lrtr_free(stream);
 		return NULL;
 	}
 
-	s->start = s->stream;
-	s->size = size;
-	s->w_head = 0;
-	s->r_head = 0;
-	return s;
+	stream->start = stream->stream;
+	stream->size = size;
+	stream->w_head = 0;
+	stream->r_head = 0;
+	return stream;
 }
 
 /* cppcheck-suppress unusedFunction */
 struct stream *copy_stream(struct stream *s)
 {
 	struct stream *cpy = init_stream(s->size);
+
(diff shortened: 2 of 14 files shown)


Brias:
@tanneberger I checked the code again and saw that at least for `init_stream` in the `bgpsec` module the return value is never checked for `NULL` and thus the newly introduce NULL-check will move the NULL-pointer dereferencing one level up. I will add more changes to ensure that error codes / `NULL` returns are handled properly in the calling code as well.

rtrlib/bgpsec/bgpsec_utils.c
```diff
@@ -27,7 +27,17 @@ struct stream *init_stream(uint16_t size)
 {
 	struct stream *s = lrtr_calloc(sizeof(struct stream), 1);
```
tanneberger, line 28: maybe while you are at it. 

please rename this variable, single character variables are a clear code smell.

rtrlib/aspa/aspa.c
```diff
@@ -32,6 +32,11 @@ static enum aspa_status aspa_table_notify_clients(struct aspa_table *aspa_table,
 			size_t size = sizeof(uint32_t) * record->provider_count;
 
 			rec.provider_asns = lrtr_malloc(size);
+
+			if (!rec.provider_asns) {
```
tanneberger, line 36: I much prefer, the more verbose variant of `if(rec.provider_asns == NULL)`, but this is a personal preference.

rtrlib/bgpsec/bgpsec_utils.c
```diff
@@ -27,7 +27,17 @@ struct stream *init_stream(uint16_t size)
...
 
+	if (!s) {
+		return NULL;
+	}
+
 	s->stream = lrtr_calloc(size, 1);
+
+	if (!s->stream) {
+		lrtr_free(s);
```
tanneberger, line 37: dont you also have to free `s->stream` and potentially other pointers inside the `s` struct.

Brias: At this point only one allocation in [line 28](https://github.com/rtrlib/rtrlib/pull/298/files/7748ed3efd0093943a5332ca242240212541b788#diff-7d7029eff35dc327e00f955180250a3190bfe79bdb200545f521cb23626071c4R28) has been performed. `s->stream` is `NULL` at that point and any other field of the struct as well. Thus, from my perspective no additional `free` is needed.

Brias:
@tanneberger I added the missing checks in the calling code, renamed the `s` variable to `stream` (not everywhere but within the functions I touched), and changed the NULL-checks to `if (var == NULL)`. Please review again.

PR #315, opened 2025-07-25
rtrlib: Make API consistent

Brias:
### Contribution description

This pull request aims to make the public API of RTRlib more consistent. The following has been implemented/changed: 

- To avoid name clashes and have a consistent API naming, every function, enum, struct, ... that is exported is prefixed with `rtr_`/`RTR_`.
- `int` return types are changed to the pertinent `enum` type if applicable.
- Change `bool` type to indicate the performed operation when notifying users about changes in the SPKI and PFX tables to specific `enum` type to improve readability and make the API specification stricter.

### Testing procedure

Since the functionality hasn't changed, no new tests were added. The existing tests can be run to quickly check for regressions:

```
mkdir build && cd build
cmake ..
make 
ctest
```

### Issues/PRs references

n/a


rtrlib/rtr_mgr.h
```diff
@@ -241,8 +242,9 @@ bool rtr_mgr_conf_in_sync(struct rtr_mgr_config *config);
  * @return PFX_SUCCESS On success.
  * @return Any other `pfx_rtvals` code depending on the error.
  */
-enum pfx_rtvals rtr_mgr_validate(struct rtr_mgr_config *config, const uint32_t asn, const struct lrtr_ip_addr *prefix,
-				 const uint8_t mask_len, enum pfxv_state *result);
+enum rtr_pfx_rtvals rtr_mgr_validate(struct rtr_mgr_config *config, const uint32_t asn,
```
tanneberger, line 245: do you only want to change the data-types and add the RTR_ prefix?

if not then, this needs to change to `rtr_mgr_roa_validate`

rtrlib/rtr_mgr.h
```diff
@@ -252,12 +254,12 @@ enum pfx_rtvals rtr_mgr_validate(struct rtr_mgr_config *config, const uint32_t a
...
  * @param[out] result The result of the AS path validation, i.e., whether the AS path is
  *                    considered valid, invalid, or whether the validation status is unknown
- * @return ASPA_SUCCESS on success
+ * @return RTR_ASPA_SUCCESS on success
  * @return Any other `aspa_status` code depending on the error.
  */
-enum aspa_status rtr_mgr_verify_as_path(struct rtr_mgr_config *config, uint32_t as_path[],
-							     size_t len, enum aspa_direction direction,
-							     enum aspa_verification_result *result);
+enum rtr_aspa_status rtr_mgr_verify_as_path(struct rtr_mgr_config *config, uint32_t as_path[], size_t len,
```
tanneberger, line 260: maybe for consistency we also should name this `rtr_mgr_aspa_validate`, this is less descriptive but more consistent with the other names.

rtrlib/rtr_mgr.h
Brias: That was my plan, but you're right that we wanted to clarify the validate function. I will change it to `rtr_mgr_roa_validate`.

rtrlib/rtr_mgr.h
```diff
@@ -269,8 +271,8 @@ enum aspa_status rtr_mgr_verify_as_path(struct rtr_mgr_config *config, uint32_t
  * @return SPKI_SUCCESS On success
  * @return SPKI_ERROR If an error occurred
  */
-int rtr_mgr_get_spki(struct rtr_mgr_config *config, const uint32_t asn, uint8_t *ski, struct spki_record **result,
-		     unsigned int *result_count);
+enum rtr_spki_rtvals rtr_mgr_get_spki(struct rtr_mgr_config *config, const uint32_t asn, uint8_t *ski,
```
tanneberger, line 274: maybe prefix this kind of functions, which subsystem they will access.

so something like: `rtr_mgr_bgpsec_get_spki`

rtrlib/rtr_mgr.h
```diff
@@ -288,7 +290,7 @@ const char *rtr_mgr_status_to_str(enum rtr_mgr_status status);
  * @param[in] fp Pointer to callback function with signature \c pfx_for_each_fp.
  * @param[in] data This parameter is forwarded to the callback function.
  */
-void rtr_mgr_for_each_ipv4_record(struct rtr_mgr_config *config, pfx_for_each_fp fp, void *data);
+void rtr_mgr_for_each_ipv4_record(struct rtr_mgr_config *config, rtr_pfx_for_each_fp fp, void *data);
```
tanneberger, line 293: maybe prefix this kind of functions, which subsystem they will access.

so something like: `rtr_mgr_roa_for_each_ipv4_record`

rtrlib/rtr_mgr.c
```diff
@@ -732,16 +739,16 @@ RTRLIB_EXPORT const char *rtr_mgr_status_to_str(enum rtr_mgr_status status)
 
 /* cppcheck-suppress unusedFunction */
 RTRLIB_EXPORT inline void rtr_mgr_for_each_ipv4_record(struct rtr_mgr_config *config,
-						       void(fp)(const struct pfx_record *, void *data), void *data)
+						       void(fp)(const struct rtr_pfx_record *, void *data), void *data)
 {
-	pfx_table_for_each_ipv4_record(config->pfx_table, fp, data);
+	rtr_pfx_table_for_each_ipv4_record(config->pfx_table, fp, data);
```
tanneberger, line 744: check if `config->pfx_table` is not `NULL`

=> also do this for the other functions that access subsysstems

tanneberger (changes requested):

rtrlib/rtr_mgr.c
Brias: How exactly do you want the check to be? `rtr_pfx_table_for_each_ipv4_record` already has an assert. Do you instead want a normal `if` check and only log that the feature is not enabled / the table is not initialized?

Changing the return type from `void` to `enum` just to indicate when the feature is not enabled doesn't make sense to me.

tanneberger: I think this is probaby a bit depending on the use-case and how often its being called. 

but for clarity & security it is probably worthwhile to actively check for it and throw and error.

tanneberger: I would change the return type.

PR #316, opened 2025-08-09
Fix memory leaks

Brias:
### Contribution description

This pull request fixes multiple memory leaks in the test cases and the RTRlib itself. All fixed memory leaks have been found by running `valgrind --leak-check=full <path-to-test-case>`.

**Note**: One memory leak still exists in `test_dynamic_groups.c`. To me it looks like it's caused by the test code and not the RTRlib and thus it can be fixed later.

**Note**: Valgrind shows memory leaks (still reachable) for the tests that start new threads. These are, however, not caused by any code in the tests or RTRlib but by the C runtime and thus can be ignored or rather ignores should be created for those leaks in the future.

### Testing procedure

Run `valgrind --leak-check=full <path-to-test-case>` for each test with and without the changes of this pull request.

### Issues/PRs references

n/a

### DIFF
--- a/rtrlib/config.h
+++ b/rtrlib/config.h
@@ -1,0 +1,24 @@
+/*
+ * This file is part of RTRlib.
+ *
+ * This file is subject to the terms and conditions of the MIT license.
+ * See the file LICENSE in the top level directory for more details.
+ *
+ * Website: http://rtrlib.realmv6.org/
+ */
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#ifndef RTR_CONFIG_H
+#define RTR_CONFIG_H
+
+#define RTRLIB_BGPSEC_ENABLED
+
+#endif
+
+#ifdef __cplusplus
+}
+#endif
+
--- a/rtrlib/rtr/rtr.c
+++ b/rtrlib/rtr/rtr.c
@@ -42,8 +42,14 @@ struct rtr_fsm_start_args {
 	struct rtr_socket *rtr_socket;
 };
 
+/** The key for registering a destructor callback that is called once a processing thread terminates. */
+static pthread_key_t processing_thread_destructor_key;
+static pthread_once_t processing_thread_destructor_key_once = PTHREAD_ONCE_INIT;
+
 static void rtr_purge_outdated_records(struct rtr_socket *rtr_socket);
 static void *rtr_fsm_start(struct rtr_fsm_start_args *args);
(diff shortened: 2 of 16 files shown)


rtrlib/config.h
```diff
@@ -0,0 +1,24 @@
+/*
```
tanneberger, line 1: why this config file?

Is this necessary, the `RTRLIB_BGPSEC_ENABLED` can also be passed via a command line to your compiler

rtrlib/rtr/rtr.c
```diff
@@ -94,12 +100,24 @@ int rtr_init(struct rtr_socket *rtr_socket, struct tr_socket *tr, struct pfx_tab
 	return RTR_SUCCESS;
 }
 
+static void make_key()
```
tanneberger, line 103: bad function name

rtrlib/config.h
Brias: Good catch. This file is in the .gitignore and shouldn't be pushed. Idk how it got committed and pushed. I will remove it again.

tests/test_live_disabled_features.c
```diff
@@ -21,7 +21,7 @@
  * (https://www.ripe.net/analyse/internet-measurements/
  *  routing-information-service-ris/current-ris-routing-beacons)
  */
-const int connection_timeout = 20;
+const int connection_timeout = 80;
```
tanneberger, line 24: why this bump?

tests/unittests/test_packets_static.c
```diff
@@ -113,7 +113,7 @@ static void test_rtr_get_pdu_type(void **state)
 static void test_pdu_to_network_byte_order(void **state)
 {
 	struct pdu_serial_query pdu_serial;
-	struct pdu_aspa *aspa = malloc(24);
+	struct pdu_aspa *aspa = rtr_malloc(24);
```
tanneberger, line 116: thats a magic number `24 == sizeof(struct pdu_aspa)` ?

tests/unittests/test_packets_static.c
```diff
@@ -162,7 +162,7 @@ static void test_pdu_to_host_byte_order(void **state)
 {
 	struct pdu_serial_notify pdu_serial;
 	struct pdu_end_of_data_v1_v2 pdu_eod;
-	struct pdu_aspa *aspa = malloc(24);
+	struct pdu_aspa *aspa = rtr_malloc(24);
```
tanneberger, line 165: same here

tanneberger (changes requested):

PR #323, opened 2026-03-14
Fix rtrclient -h exit with success code 0

fvompe:
Fix small inconsistency: help command should be successfull, because user gets help as requested.

PS. I am using it in Dockerfile to check that rtrlib was correctly compiled.

## Before

```bash
$ ./rtrclient -h; echo $?
...
1
```

## After

```bash
$ ./rtrclient -h; echo $?
...
0
```


Brias:
@fvompe Thanks for the PR. Looks good to me :+1:

Brias (approved):

PR #327, opened 2026-04-21
Cmake: Fix OpenBSD build with libssh by adding link_directories

mwinter-osr:
OpenBSD installs the libssh libs in /usr/local/lib which isn't searched with a -lssh. Need to include the LIBSSH_LIBRARY_DIRS to find the libraries on OpenBSD

Fix verified on OpenBSD 7.8
This is part of issues found on the NetDEF CI system


eqvinox:
superseded by #330

Brias:
Not needed since the changes have been merged by #330

PR #330, opened 2026-05-20
cmake: OpenBSD, but actually working

eqvinox:
#327 gets OpenBSD mostly working, but the way the library is installed the linker won't find it due to OpenBSD doing its own thing with versioning

(also for link_directories there seems to be a more "modern" cmake way to do it)

Also, building FRR against it fails due to `aspa.h` including the internal `alloc_utils_private.h` file.


Brias (approved):
Looks good. Thank you 👍

